PR #231, opened 2026-07-30
test(velvet): pin the contributor README against the sources it names

### LINKED ISSUE #222: Generators~/README.md carries unpinned facts and is outside the documentation drift guard
`DocumentationDriftTests` resolves paths and identifiers for `Documentation~/*.md`, both `README.md` files and `CLAUDE.md`. `Generators~/README.md` is not in that set, and it is where the analyzer diagnostics' definitions live — what counts as a level of nesting, what counts as a branch, what counts as a demanded argument — plus the mutation-testing guidance.

Two changes have already had to state that nothing mechanically pins that prose.

### What adding it would flag today

Measured against the current guard's rules: **0 unresolved paths**, and 5 unresolved identifiers — `Analyzers`, `RoslynAdditionalFileImporter`, `USS001`, `USS011`, `VEL`. All five are real (a directory name, a Unity importer type, USS diagnostic IDs, and an ID prefix written bare), so they want an allowlist entry each with a stated reason, in the three-group form the fixture already uses.

### The part the guard cannot reach

The identifier check pins names, not numbers. `Generators~/README.md` also carries figures — a mutation score, a per-project table — which rot on the next commit and which nothing checks. Adding the file to the guard does not address that; deciding whether those figures belong in a document at all is the separate question, and the mutation-testing change already dropped the percentage column for exactly this reason.

### Acceptance criteria

- `Generators~/README.md` is in `TargetMarkdownFiles`, with the five allowlist entries justified individually rather than as a block.
- The `Documentation~/README.md` index check is confirmed not to apply to it, or is extended deliberately.
- A statement of which facts in that file remain unpinned after the change, so the next person does not read inclusion as coverage.

s4k10503:
`Packages/com.velvet.core/Generators~/README.md` states the analyzer definitions, the paths of the three committed artifacts and the layout of the generator test project, and nothing checked any of it — two earlier changes had to say so in their own descriptions. `DocumentationDriftTests` now scans it alongside the `Documentation~` guides, both `README.md` files and `CLAUDE.md`.

## What adding it reported, and what each is

Measured on the tree at 1c5258c: **0 unresolved paths, 12 unresolved identifier findings over 11 distinct names.**

| name | disposition |
| --- | --- |
| `Analyzers` (twice) | a real directory, reached through `../Runtime/Plugins/Analyzers/`. The path check refused the span for its leading `../` and let it fall through to the identifier check. Fixed in the guard, not allowlisted. |
| `DOTNET_ROOT`, `StrykerOutput` | an environment variable the .NET apphost reads and the directory the mutation runner writes — an external toolchain the file states how to invoke, not something this repo holds. |
| `RoslynAdditionalFileImporter` | a Unity importer type, the same group as the `RoslynAnalyzer` label already in the list. |
| `USS001`, `USS011`, `VEL500`, `VEL501`, `VEL502` | real diagnostic identifiers that C# holds only as string literals, which this fixture's corpus strips by design. Allowlisting them forfeits nothing: `DocumentationDiagnosticTableTests` in the `Generators~` suite reads this same README and compares its `VEL` and `USS` spellings against the real descriptors and the derivation's real code range. |
| `VEL` | the ID space written as a shape (`VEL###`), not an ID — the `Generators~` guard's `VEL\d{3}` pattern is right not to match it. |
| `Shape` | the analyzer category `Velvet.Shape`, also a bare string literal, and pinned by nothing on either side. |

The relative-path fix is load-bearing rather than cosmetic. Allowlisting `Analyzers` instead would have silenced a whole class of real drift: a path whose every component is still a live word somewhere — a committed artifact emitted to a moved directory, say — passes the identifier check untouched, so the file's existence would never be asked about. The path check now strips a leading `./` or `../` run and matches the remainder as a suffix, exactly as it already does for every other reference.

`Documentation~/README.md`'s index check is unaffected: it enumerates the `Documentation~` directory, and this file is not in it, so it is neither expected in the index nor reported as a dead link.

## What still is not pinned in that file

Inclusion is not coverage. After this change, nothing checks:

- anything inside a fenced block — the whole directory-layout tree is stripped before spans are read;
- markdown link targets, since only backticked spans are scanned;
- the mutation-testing figures (mutant totals, the CodeFixes breakdown, the wall-clock comparisons), which rot on the next commit;
- the three code-shape limits stated in prose (4 levels, 20 branches, 6 parameters) against `NestingDepthAnalyzer.MaxDepth`, `BranchCountAnalyzer.MaxBranches` and `ParameterCountAnalyzer.MaxParameters`;
- the analyzer category name;
- whether any description is still accurate, which stays the author's job.

No `Documentation~` guide describes this fixture, and `CLAUDE.md`'s sentence about what the drift guards catch stays true as written; the fixture's own summary is updated to name the file it now reads. No CHANGELOG entry: contributor tooling, with no user-visible behaviour change.

Closes #222.

The allowlist comment runs past the twelve-line review trigger deliberately. Its load-bearing facts are the two the table above turns on: that the diagnostic identifiers are checked by the `Generators~` suite rather than here, and that `VEL` and `Shape` are checked by nobody.

### DIFF
--- a/Packages/com.velvet.core/Runtime/Component/Tests/Editor/DocumentationDriftTests.cs
+++ b/Packages/com.velvet.core/Runtime/Component/Tests/Editor/DocumentationDriftTests.cs
@@ -9,8 +9,9 @@ using NUnit.Framework;
 namespace Velvet.Tests
 {
     /// <summary>
-    /// Machine-checks the shipped Documentation~ guides (plus both README.md files and CLAUDE.md) against the actual
-    /// runtime API surface, so a doc referencing a renamed/removed <c>V.*</c> factory or <c>Hooks.*</c> hook,
+    /// Machine-checks the shipped Documentation~ guides (plus both README.md files, CLAUDE.md and the
+    /// contributor README under Generators~) against the actual runtime API surface, so a doc referencing
+    /// a renamed/removed <c>V.*</c> factory or <c>Hooks.*</c> hook,
     /// a path or a type that no longer exists, or an index that has drifted from the files on disk, fails a
     /// test instead of shipping silently wrong. Each check pins a failure mode that has actually shipped: a
     /// guide referencing a never-implemented factory, a hook table drifting from the real hook surface, an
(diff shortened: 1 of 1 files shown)


s4k10503:
The relative-path call is right, and fixing it in the guard rather than allowlisting `Analyzers` is the part that mattered — the reasoning about a path whose every component is still a live word is exactly the failure that entry would have hidden.

One entry is worth a second look: **`Shape`**. The allowlist is global across every scanned document, so admitting a word that common means any future document can write `Shape` in a backticked span and be believed, including one naming a real type that has since been renamed away. The other generic entries already present (`Move`, `Leave`, `Up`, `Wheel`) came in as a named group — Unity event names — which at least bounds what they are excusing; `Shape` is admitted for one occurrence in one file, and pays for it everywhere.

It is a half of `Velvet.Shape` that the scanner splits, so the alternatives are to make the category a checked fact on the `Generators~` side (where the descriptors already are), or to leave the README naming the category in a form the scanner does not split. Either is better than a permanent global exemption for the word. Your own note says nothing pins it, which is the same conclusion from the other direction.

Not blocking if you disagree with the trade — say which and it goes in as is.

s4k10503:
Verified the guard independently of the report, on `00ba83f`.

Renaming one committed-artifact path in `Generators~/README.md` — `../Runtime/Styling/StyleUtilityProperties.g.cs` to a name that does not exist — turns `DocumentationDriftTests` red: **4 passed, 1 failed**, and the failure message names the injected path. Reverting it: **5 passed, 0 failed, 0 inconclusive**.

That is the specific thing worth confirming here, because the `../` prefix is what the path check previously refused. Before this change the same rename would have fallen through to the identifier check and passed on `Runtime`, `Styling` and `StyleUtilityPropertiesRenamed` each being resolvable words — which is the failure the PR body argues about, now shown rather than argued.

EditMode only, and only this fixture; the full-suite result is yours.

PR #233, opened 2026-07-30
chore(velvet): hold the generator solution to the code-shape rules it implements

### LINKED ISSUE #223: Generators~/src violates both code-shape rules it enforces
VEL500 (nesting depth > 4) and VEL501 (branch count > 20) are error severity on every package assembly that carries the opt-in marker. `Generators~` is a standalone .NET solution that never references Velvet, so the analyzers never load there, and the backlog guard excludes it deliberately.

Measured with the analyzers' own logic against that source:

- `MemoizeMethodGenerator.BuildCandidate` — **27 branches**
- `Purity/SideEffectWalker.IsImmutableCapture` — **23 branches**

No document claims the repository obeys its own limits everywhere, so nothing shipped is false. But "the repo obeys its own limits" is not true of `Generators~`, and the code that *implements* the rules is the least obvious place to leave exempt.

### Options, in the order I would consider them

1. **Reference the analyzer package from the three `Generators~` csprojs** and drain. The analyzers are built there, so the wiring is short — but a project that analyses itself with its own output needs its build order checked.
2. **Extend `CodeShapeBacklogDriftTests` to that source tree** without loading the analyzer into its compile. The guard already parses package sources from a dotnet process, so this is the cheaper half: it catches new violations without requiring the drain first.
3. **Document the exemption** and stop there. Cheapest, and the weakest — an unstated exemption is what this issue is.

### Acceptance criteria

- Either the two violations are drained and new ones are mechanically caught, or the exemption is stated where a reader of the rules will meet it.
- If drained, the same order applies as for the package: warning, measure the real backlog with the analyzer itself, drain to zero, then promote.
- A note on whether `Generators~` should share the package's limits at all — it is a different kind of code with a different reader, and "the same numbers" may be the wrong answer even if "no limit" also is.

s4k10503:
Closes #223.

## The exemption

`VEL500`, `VEL501` and `VEL502` were errors on every assembly under `Packages/com.velvet.core/` and on none of `Generators~`, which is where all three are written. That solution never references Velvet, so the analyzers never loaded there, and `CodeShapeBacklogDriftTests` skipped the tree deliberately.

Every project in the solution now declares the opt-in marker and loads the analyzers from `Velvet.SourceGenerators.Bootstrap`, a second compile of the analyzer sources under a second assembly name.

Two cheaper wirings were tried first and do not work:

- A project referencing **itself** as an analyzer fails during restore, before any compile: `NuGet.targets(1294,5): error MSB4006`, a circular dependency in the target dependency graph involving `_GenerateRestoreProjectPathWalk`. That is what rules out wiring `Velvet.SourceGenerators` — the project holding three of the four violations — without a second compile.
- Pointing `<Analyzer Include=...>` at the committed `Runtime/Plugins/Generators/Velvet.SourceGenerators.dll` avoids the second compile, but nothing verifies that DLL was rebuilt from the current sources. A change to a rule would then be enforced by the previous rule until someone remembered to redeploy — the same "remember to do it" failure the wiring exists to remove.

## The backlog

With the wiring in place the compiler reported four members over a limit — one more than #223 listed, and one of them under a different rule:

| member | file | measured | limit |
| --- | --- | --- | --- |
| `MemoizeMethodGenerator.BuildCandidate` | `src/Velvet.SourceGenerators/MemoizeMethodGenerator.cs` | 27 branches | 20 |
| `SideEffectWalker.IsImmutableCapture` | `src/Velvet.SourceGenerators/Purity/SideEffectWalker.cs` | 23 branches | 20 |
| `CodeShapeMembers.BodiesOf` | `src/Velvet.SourceGenerators/CodeShape/CodeShapeMembers.cs` | 21 branches | 20 |
| `StyleUtilityTableBuilder.CollectRule` | `src/Velvet.StyleTable/StyleUtilityTableBuilder.cs` | 7 parameters | 6 |

Opting the test project in surfaced a fifth: `CodeShapeBacklogDriftTests.Measure`, nesting 5 levels.

Each is drained by decomposition rather than by moving branches somewhere they stop being counted:

- `BuildCandidate` splits along the axis its diagnostics already divide it by — the requirements on how the method and its enclosing types are *written*, the requirements on its *signature*, and the deps-less purity warning that sits between the two because it is suppressed by the first and not by the second. The emitted diagnostic sequence is unchanged: `VEL001` and `VEL002` are mutually exclusive, so the arity check moving into the signature half cannot reorder anything.
- `IsImmutableCapture` states the sixteen built-in immutable types as a set rather than as a sixteen-arm switch.
- `BodiesOf` collapses its five method-like cases into one. They find their body identically and differ only in the token a report points at, which is now a single expression.
- `CollectRule` takes the derivation state as one accumulator instead of four values threaded three calls deep. `StyleUtilityProperties.g.cs` regenerates byte-identical.
- `Measure` lifts its per-node work into one recorder per measured surface.

## Keeping it mechanical

The wiring alone leaves a second-order hole one level up: enforcement reaches each project through two MSBuild items, and a project added later carrying neither compiles clean and is exempt with nothing anywhere to say so — the state the whole solution was in. `GeneratorProjectOptInDriftTests` fails when a project under `Generators~` lacks either. Only one of the two halves leaves a trace in metadata, so both are read out of the project files.

A fourth case compares the bootstrap's declared types against the project it stands in for. Its sources arrive through a glob into the sibling's directory, and a glob that stopped matching would leave an analyzer assembly holding no analyzers — every project would then compile clean with nothing enforced and nothing failing.

### The hole this leaves

Solution membership is not checked. A project carrying both MSBuild items but missing from `Velvet.SourceGenerators.sln` is never compiled, so the rules never run on it — the same silent exemption this change set out to close, one level up. That guard is being added under #243 and is deliberately not here, so the two do not collide.

The escape is not silent, which is what this paragraph claimed twice and measurement refuted twice. Three of the four guards stay green; the marker guard reads a **built** assembly, so an unbuilt project makes it throw. It now throws naming the likely cause — a project absent from the solution is not compiled — rather than reporting a missing DLL, which sent the reader after a build failure instead of after the missing entry.

## Should `Generators~` share the package's numbers?

Yes, and #223 is right to ask. The limits are not about what the code does but about how much of a member a reader must hold at once, and that budget does not change with the kind of code. If anything the asymmetry runs the other way: the analyzer sources are read cold — by someone who arrived to change one rule and has no model of the file — more often than the runtime is.

The one shape that could have argued for different numbers is a dense parser or dispatcher, where width is the problem rather than a symptom of it. All four members were that shape, and all four came apart along a seam already present in them. That is weak evidence for one set of numbers rather than two, but it is the evidence this change produced.

No CHANGELOG entry. Nothing a consumer can observe changes: the analyzers' diagnostics, the generator's output and the derived style table are identical, and the marker now carried by the generator assemblies means nothing to a compile that only loads them.

`Generators~/README.md` owns the build and the layout, so it gains a **How this solution opts in** section: the two MSBuild items a project needs, why the analyzers arrive from a second compile, and what the glob is load-bearing for. Its layout tree lists the new project.

Two statements already in that file were false and are corrected rather than left in place:

- `Velvet.SourceGenerators.csproj` claimed its output is byte-identical across rebuilds. It is not — `ContinuousIntegrationBuild` embeds the git SHA in `AssemblyInformationalVersion`, so the committed DLL holds the SHA of the commit *before* the one carrying it and no rebuild can reproduce those bytes. This is the measured reason the "point the analyzer at the committed DLL" option is unsafe rather than merely inelegant, and it means nothing can check the deployed artifacts against a fresh build.
- The layout tree said `Velvet.StyleTable.csproj` was force-added because the root `.gitignore` ignores `*.csproj`. That pattern is `/*.csproj`, anchored to the repository root, so it never matched any project file under `Generators~`; `git check-ignore` exits 1 on all of them, and the four `!` negations in `.gitignore` that exist to counter it are dead.

`MSB4006` joins `DocumentationDriftTests`' identifier allowlist beside `DOTNET_ROOT` and `StrykerOutput` — an error code MSBuild owns, which no source file here spells. The drift guard caught it; that is what found this.


s4k10503:
Handled the two items left to me.

**The `-s ours` merge needed nothing.** `41891e7` is already on the branch and `origin/main` has no commit missing from it. The squash diff is 16 files with no deletion of anything main added, so the flattened tree is the one intended — checked rather than assumed, since a `-s ours` merge is exactly the operation that can silently discard the other side.

**The mutation counts are withdrawn rather than re-measured.** They were taken before this PR decomposed the four mutated files, so they described a tree that no longer exists. The section already made the argument against quoting them — a figure that is stale by the next commit and that nothing re-derives — and then quoted the absolute counts anyway. The `CodeFixes` figures stay, dated to the commit they came from and marked not re-measured, because what they support is that the project is covered by almost nothing: a shape rather than a size, and nothing since has aimed at it.

Re-measuring was the alternative and was rejected on proportion: a full run is around half an hour, no job gates on it, and the fresh numbers would rot on the next edit to a mutated file — which is the property that made the old ones wrong.

`DocumentationDriftTests` 5 passed / 0 failed, generator suite 383/383 after the edit.

PR #237, opened 2026-07-30
test(velvet): fold the assume gates left in three fixtures

### LINKED ISSUE #221: Assume statements still gate the behaviour under test in three fixtures
An `Assume` that gates the behaviour a test exists to pin converts a regression into an Inconclusive, which the NUnit runner does not count as a failure — only `scripts/assert-no-inconclusive.py` surfaces it. Twenty-five such statements were cleared in #216 and thirty-five in #208; these are what the same sweeps left behind.

**`ClipPathWrapTests`, 9 statements** in the clip-path half. Untouched by #218, which rewrote the ring half of the same file.

**`ShadowAnimationVisibilityTests`, 2 statements** — `resolvedStyle.opacity < 1f` and `> 0f`. If the spring never ran, the real assertion would fail; the `Assume` converts that into Inconclusive.

**`StyleRingClassTests.Extract`** wraps `TryExtract`, so roughly 20 tests report Inconclusive rather than Failed if extraction breaks. This one is defensible under the convention — it states a fact about a separately-tested component — so it may be correct to leave. Decide explicitly rather than by omission.

### The technique, established

Neuter the mechanism, record which tests report Inconclusive, then decide fold-versus-delete **per statement using the neuter rather than by reading**. On the z-layer sweep the inference "already Failed means it was not gating" held for exactly 1 of 10 — the other nine threw before reaching the `Assume` line, which says nothing. Read failure messages, not counts.

Two findings from #216 worth carrying:

- **The logically sharpest gate is not always the one that discriminates.** A `ReferenceEquals` precondition looked like the load-bearing one, but folding only it left the test green, because the pool is LIFO and returns the same instance anyway. The pool-count term was what made it red.
- **NUnit's tuple comparison matches a nested collection by reference.** A fold comparing two string arrays inside a tuple looked correct and was not. It surfaced only from a per-fixture breakdown; a check that N results moved from Inconclusive to Failed would have passed it through.

### Acceptance criteria

- Each statement is folded into its assertion as a single tuple comparison, or deleted where the assertion alone still fails on the broken behaviour — with the neuter showing which.
- Inconclusive becomes Failed under the neuter, verified per test rather than by count.
- Any case that still passes under its neuter after folding is named as a hole, even where closing it needs a structural change that is not made.

s4k10503:
Closes #221.

## What changed

Fifteen `Assume` statements that gated the behaviour their case exists to pin. An `Assume` failure reports Inconclusive, which the NUnit runner does not count as a failure — only `scripts/assert-no-inconclusive.py` surfaces it, so a regression behind one is invisible in a green run.

The convention was applied rather than followed, so the fifteen went three ways:

- **Thirteen folded** into their assertion as a single tuple comparison — eight in `ClipPathWrapTests`, five in `ShadowAnimationVisibilityTests`. Each keeps one assert per case.
- **One deleted**, in `ClipPathWrapTests.Given_AUserWrapElementWithClipClass_When_Patched_Then_NotDoubleWrapped`. A first pass folded the `Assume` in as the **first** term, giving `(boundAtCreate, ClipPathBindings.Count)` against `(0, 0)`. Measurement showed that first term was redundant: a create path ignoring the opt-out leaves the binding in place for the post-patch count to find, so under that neuter the pair read `(1, 1)` and the second term — the original assertion — already goes red on its own (`Expected: 0 But was: 1`). The folded-in term was dropped and the assertion left as it was. This is the convention's delete branch: deletion is correct exactly when the assertion alone still fails on the broken behaviour.
- **One replaced**, in `StyleRingClassTests`. Its `Extract` helper wrapped `TryExtract` in an `Assume` — not a fact about a separately-tested component but about the parser the fixture exists to pin. A `TryExtract` hardwired to return false left sixteen cases inconclusive and **none failed**, which the runner reports as a pass. A failed extraction now resolves to a sentinel spec no assertion can match, so each case fails on its own and keeps its single assert.

Fourteen of the fifteen were confirmed gating by neutering the mechanism they depend on; all twenty-nine affected cases moved Inconclusive → Failed, checked per case against failure text rather than by count. The fifteenth — the spring-exit precondition in `ShadowAnimationVisibilityTests` — could not be made to fire at all: the simulated panel loads no stylesheet, so `opacity-0` never resolves to zero. It was folded into a bracket that does discriminate.

## Two claims in the issue that measurement contradicted

The issue described the spring pair as "if the spring never ran, the real assertion would fail; the `Assume` converts that into Inconclusive". Measured, it is the other way round: with the spring's tick never scheduled the exit case reported **Failed**, and the enter case **passed outright**, because an enter parks both the caster and its shadow at 0 and "below full" accepts both. Both now bracket the two values strictly inside the range an enter or exit travels. The issue also counted two statements in that fixture; there are five.

## Known holes, with the cut each was measured at

A case asserting only that something is *absent* also holds against a feature that does nothing. These were enumerated by running every case that references a feature against that feature killed at each layer and diffing, because a count taken by reading moved every time it was re-checked — a case whose expected warning or whose suppression gate reads the *parser* stays green when only the *applier* is neutered.

- Clip: five vacuous at the applier cut, three at the parser cut.
- Ring, reconciler fixture: four at the applier cut.
- Ring, parser fixture: the five `Is.False` extract cases plus the `HasRingClass` gate case — six, the same shape one method over.

Closing any of them needs a control that *resolves* in the same assertion, which is a fixture restructure rather than a fold, so none is made here. They are #245.

## One defect this branch introduces, which merge order removes

Three `.Within()` sites in `ShadowAnimationVisibilityTests` are inert: Unity's NUnit ships no `ValueTupleComparer`, so a tolerance never reaches a tuple's members and the comparison becomes bit-exact while still printing the tolerance it did not apply. Those three had a working scalar tolerance before the fold and lost it in the conversion — a cost of the fold convention itself, not of these cases. They pass only because their operands are literal constants. **#236 deletes that file and merges first**, so this never reaches `main`; the general rule now lives in the `unity-tests` skill, and the eight other sites are #241.

## Verification

EditMode 3541 total, 3538 passed, 0 failed, 0 inconclusive, 3 pre-existing `[Ignore]`s. PlayMode 128/128. `scripts/assert-no-inconclusive.py` clean over both. No `-nographics`; measured on a quiet machine.

No changes. Test-only, no user-visible behaviour and no API surface; `DocumentationDriftTests` passes.


s4k10503:
Round-one findings are addressed on `47bc4b9`, as a follow-up commit rather than an amend so the review anchor survives.

The comment that overclaimed is resolved by **deleting** the term rather than rewriting the claim — the author's own neuter output had already shown the second term reading `1` on its own, so the post-patch count catches a create-path fault unaided. That makes it the convention's delete branch, not its fold branch, and re-running the neuter against the single assertion confirms nothing was lost.

The five ring holes are named in the fixture, with the finding that the sentinel **structurally cannot** reach them: it substitutes for a helper's return value and those cases never call the helper. Routing them through it would assert "the spec is the no-ring sentinel", which restates "the parser said no" and a dead parser satisfies equally. They need a positive control each — a restructure, not a fold — and are #245 along with the four clip cases.

The count is now stated with the cut it was taken at: **four at the applier cut, three at the parser cut.** `Given_ClipOnMotion` differs between them because its `LogAssert` expectation is gated on `WantsClipPath`, which the applier cut leaves intact — so the log expectation is satisfied by the parser while `Is.False` passes for want of a wrapper. The earlier reasoning that the `LogAssert` protected it was true only at the parser cut, and was cited without that qualification.

Full suites on `8fa7fd3`: EditMode 3538 passed / 0 failed / 0 inconclusive (3 pre-existing `[Ignore]`s), PlayMode 128/128, `assert-no-inconclusive.py` clean on both.

Still held for sequencing: no rebase, and `ShadowAnimationVisibilityTests.cs` untouched, pending #236 which deletes that file.

PR #239, opened 2026-07-30
feat(velvet): resolve the bundled utility stylesheet from runtime code

### LINKED ISSUE #227: No supported path gets the bundled stylesheet onto a panel in a player build
## What is wrong

`Runtime/Styles/StyleUtilities.uss` carries every utility class the framework's styling layer is built on. Two places in the repository resolve it, and neither exists in a build:

- `TestUtilities/StyleUtilitiesTestLoader.cs` — `AssetDatabase.LoadAssetAtPath`, wrapped in `#if UNITY_EDITOR`, and its own summary says "Test-only. Must not be used from production code."
- `Editor/Preview/VelvetPreviewWindow.cs` — the same call from the `UnityEditor` assembly.

`Runtime/Preview/VelvetPreviewHost.cs` attaches a sheet but does not resolve one: it consumes whatever `VelvetStyleHints.PreviewStyleSheet` was set to by its caller.

So no public runtime API returns or attaches the bundled sheet. A user who installs the package, mounts a tree, and builds gets an unstyled panel — every plain utility class silently does nothing, while arbitrary-value utilities keep working because those resolve to inline style. That is the same failure mode that has already produced wrong conclusions inside this repository's own pixel fixtures, and a user hitting it has no test harness to reveal it.

## What a user has to do instead

Reference the `.uss` from a serialized field or a `PanelSettings`/UXML asset so the build pipeline pulls the package asset in. That works and is ordinary Unity practice, but it is stated in no document: `Documentation~` contains no setup page, and neither README mentions `styleSheets` or the asset path.

## Acceptance criteria

- A supported way exists to get the bundled utilities onto a panel from runtime code in a player build, or — if the deliberate answer is that the user must reference the asset from the scene — that is stated as the supported path.
- Whichever it is, decide it by measurement in a built player, not from source: confirm the sheet's rules actually apply after the chosen path, using a plain (non-arbitrary-value) utility class, since an arbitrary-value one passes either way.
- If a runtime resolver is added, it must not fork the stylesheet's location: `Generators~/src/Velvet.StyleTable` reads `Runtime/Styles/*.uss` to emit `Runtime/Styling/StyleUtilityProperties.g.cs`, and the test loader and preview window hold the path as a literal. Any move updates all of them, and the drift guard should pin the result.
- A setup document under `Documentation~` states it, listed in `Documentation~/README.md`, and the package `README.md` links to it from the install section.

Related: #226 (the package ships nothing runnable, so this has never been exercised end to end).

s4k10503:
Closes #227.

## What was wrong

Every plain utility class Velvet ships is a USS rule in `Runtime/Styles/StyleUtilities.uss`. Both resolvers of that sheet went through `AssetDatabase`, which does not exist in a player, so no public API could put it on a panel in a build. A shipped game resolved every plain class to nothing while arbitrary-value classes such as `w-[120px]` kept working, because Velvet turns those into inline style — a missing sheet that presents as a partial styling bug. No document said what to do instead.

## What this adds

`VelvetStyleUtilities.AttachTo(root)` attaches the sheet, in the editor and in a player alike; `VelvetStyleUtilities.Sheet` returns the asset. It resolves through a one-line `Resources` asset, `Runtime/Resources/Velvet/StyleUtilities.uss`, that `@import`s the real sheet rather than replacing it — so the sheet keeps the single location the style-table generator and every existing literal already point at, and the import pulls it and every partial into the build as dependencies.

The cost, stated plainly: the package's `Resources` folder is included in every build of every project with the package installed, whether or not anything calls `AttachTo`.

The scene-reference alternative (a serialized `StyleSheet` field, `PanelSettings`, or UXML) still works and needs no runtime code; `Documentation~/setup.md` covers both and says when to prefer which.

## Test-side consequence

The shared editor-only loader in `TestUtilities` was `#if UNITY_EDITOR`, and two PlayMode fixtures called it — so `Velvet.Tests.Styling.PlayMode` could not compile into a player at all, and no PlayMode fixture had ever exercised a styled panel outside the editor. That loader is gone; its call sites and the preview window now use the same public API a user calls.

## Evidence

Measured in a **built macOS standalone player** (`-testPlatform StandaloneOSX`), which is where `AssetDatabase` is absent: `BundledStyleUtilitiesRuntimeTests` mounts `flex-row` on a `UIDocument` panel and reads `resolvedStyle.flexDirection`. Without the attach it is `Column`; with it, `Row`. A plain class was chosen deliberately — an arbitrary-value class passes either way.

Pointing the `Resources` asset's `@import` at a rule-less partial makes only the attached case fail (`Expected: Row / But was: Column`) while the sheet still loads, which is what shows the assertion is measuring the sheet rather than the scaffolding.

`RuntimeStyleSheetResourceTests` in the `Generators~` suite pins the import target against the aggregator's real location, with the file's existence folded into the same assertion so two stale paths cannot agree. It runs in the license-free workflow, which fires on any change under `Runtime/`.

## Documentation

`Documentation~/setup.md` is new, listed in `Documentation~/README.md`, and linked from the install sections of both READMEs. Both getting-started samples now attach the sheet before mounting. `CHANGELOG.md` has an `[Unreleased]` entry — this is user-visible behaviour.


s4k10503:
Round-one findings are addressed on `2b328ed`, all four checks green.

Two things about how the two MEDIUMs were resolved are worth recording, because in both cases the fix went further than the finding.

**The false claim was verified false rather than softened.** Zero `.gap-*`, `.space-*`, `.divide-*` or `ring-*` rules exist across all seventeen bundled sheets, so the restatement names the four sheet-independent families rather than hedging the absolute. It is now pinned by a case mounting `gap-4` on a sheet-less panel, red without the class at `Expected: 16.0f / But was: 0.0f` — the claim is checked instead of asserted.

**The inert test was inert further than the review said.** With the guard deleted the result is `count=2 first=True`: Unity's `Add` neither duplicates nor reorders, so nothing at all distinguishes the guarded version from the unguarded one. The right conclusion was to remove both the test and the guard it was mirroring, leaving the measured engine fact as a comment, rather than to write a test that could tell them apart — there is nothing to tell apart.

`ResourcePath` is no longer a mirror: the `Generators~` guard parses it from source, and the duplicate literal is gone.

Verified: EditMode 3538/0/0 (3 pre-existing ignores), PlayMode 132/0/0, `Generators~` 381/0, and 4/0/0 in a built player.

Separately — the `Shader.Find` gap surfaced during this work is now #246, and it is severe: all four Velvet shaders resolve in editor Play Mode and return null in a build, so drop shadows, gradient silhouettes and the shader-backed filters silently render nothing in every shipped player. Same class as the issue this PR closes, one layer down. It is out of scope here.

I misattributed that finding to this PR's author in an earlier message; it came from the review side.

Issue #243, opened 2026-07-30
Nothing mechanical catches a test that cannot fail


s4k10503:
Every defect found in the current round of review was found by a person or an agent reading code. Not one was caught by a test, a guard, or CI. Review is not repeatable and does not scale, and the defects were not exotic:

- An assertion comparing a resolved layout against the same two locals written to `style.width`/`height` a few lines above. It cannot fail in any state.
- A count term comparing the number of iterations of a loop against the length of the list the loop walks. Also cannot fail — and it shipped in the same commit that removed the first one *for being tautological*.
- Ten assertions carrying `.Within()` on a `ValueTuple`, where the tolerance is silently dropped (#241).
- A fixture reporting zero failures with the mechanism under test neutered.
- A probe attached to a panel root that changed the logical child count, so `only:` / `last:` / `nth-last-child` resolved differently than in the window the capture claims to mirror.

The suite was green through all of it, which is the point: these are not bugs the suite missed, they are **assertions that had stopped asking anything**, and a green suite is exactly what that produces.

## What is mechanically catchable today

Two of these are certain and cheap, and both should be built rather than remembered:

1. **`.Within()` applied to a tuple comparison.** Purely syntactic — `Is.EqualTo(<tuple>)` followed by `.Within(...)`. The repository already has the analyzer infrastructure (`Generators~/src/Velvet.SourceGenerators/CodeShape/`, VEL500–502) and the opt-in marker mechanism. This closes #241's class permanently instead of auditing it once.
2. **A project present on disk but absent from `Velvet.SourceGenerators.sln`.** Such a project is never compiled, so its analyzers never run, and the four guards added for exactly this purpose all stay green. Compare `*.csproj` on disk against solution membership.

## What is not, and the question that was never asked

Everything else here is one thing: a test that passes whether or not the code under test works. Mutation testing is the only general instrument for it, and #202 concluded it was infeasible for the Unity assemblies — Stryker rebuilds through an MSBuild graph, Unity compiles asmdefs in-editor with this package's ILPP in the pipeline, so every mutant costs an editor recompile plus a suite run with nothing to amortise it against.

That conclusion is about a **full** run over the whole package. It does not answer the question a pull request actually needs, which is narrower:

> Given the files this branch changed, does any test fail when each of them is mutated?

[...]

Issue #241, opened 2026-07-30
Eight tuple assertions carry a tolerance that is never applied


s4k10503:
`.Within()` does not reach the members of a `ValueTuple`. Unity's NUnit ships no `ValueTupleComparer`, so `NUnitEqualityComparer.AreEqual` falls through to `FirstImplementsIEquatableOfSecond`, which never receives the `ref Tolerance`. Running

```
Assert.That((0.99999f, 0.00001f), Is.EqualTo((1f, 0f)).Within(1e-4f))
```

under mono against `com.unity.ext.nunit` **fails**, while printing the tolerance it did not apply — which is how two authors read a passing assertion as proof the tolerance was working.

**Eight sites**, enumerated by the VEL503 analyzer in #247 rather than by grep:

| file | lines |
|---|---|
| `Runtime/Styling/Tests/Editor/ArbitraryValueTests.cs` | 1802, 1822, 1842, 1866 |
| `Runtime/Styling/Tests/Editor/SkewChildTranslateManipulatorTests.cs` | 284, 368 |
| `Runtime/Reconciler/Tests/Editor/MotionVariantPropagationTests.cs` | 293, 303 |

Each is bit-exact equality wearing a tolerance suffix, surviving only because its values happen to land exactly. The convention that produced them is sound — folding a precondition and its subject into one comparison is how this repository stops an `Assume` from turning a regression into an Inconclusive — but a scalar assertion that carried a working tolerance loses it in the fold, silently.

The sharp end is `SkewChildTranslateManipulatorTests.cs:284` and `:368`, which compare a **measured layout value** exactly. Font metrics and layout land differently on a CI runner than on a developer machine, which is precisely the case the tolerance was there for.

**This issue said ten.** That count was taken with a grep for `Within`, which also matched the identifiers `GroupFocusWithin` and `PeerFocusWithin` in `StyleVariantClassTests` — two lines carrying no tolerance at all. The analyzer finds every real site and no false one, which is the argument for enumerating with it rather than by hand.

## Acceptance criteria

- Every one of the eight either rounds each term before comparing, or compares formatted strings, or is restructured so the tolerance applies. State which, per site.
- Prove one of them would have failed before the fix and passes after, using a value inside the intended tolerance and outside exact equality — otherwise the fix is unverified in exactly the way the defect is.
[...]

PR #247, opened 2026-07-30
feat(velvet): report a tolerance that cannot reach a tuple, and a project the solution never builds

s4k10503:
Two mechanical checks from #243, for the two defect classes there that a machine can settle.

**VEL503 — a tolerance chained onto a tuple equality.** NUnit's comparer chain has no entry for `ValueTuple`, so the pair falls through to the expected value's own `IEquatable<T>`, which is never handed the `ref Tolerance`. The assertion is bit-exact equality, and its failure message still prints the tolerance it did not use — which is why two authors read a passing assertion as proof the tolerance applied. The analyzer decides on the expected value's **type** rather than its syntax, so a tuple held in a local is caught as well as a parenthesised literal.

It is a **warning**, unlike its VEL500–502 siblings, and the descriptor says why: the package's own test assemblies carry sites it reports, and an assembly that fails to compile runs none of the rules. The help text also names the two shapes it does not report — a constraint built in one statement and given its tolerance in another, and a tolerance dropped by any other expected type NUnit compares through `IEquatable<T>`.

**A project on disk but absent from the solution.** Such a project is never compiled, so its analyzers never run and every opt-in guard stays green — the same silent exemption those guards exist to close, one level up. `SolutionProjectMembershipDriftTests` compares `*.csproj` on disk against solution membership.

## Measured, and it corrects the number in #241

Repo-wide, the analyzer reports **8 distinct sites in 3 files**:

| file | sites |
|---|---|
| `ArbitraryValueTests.cs` | 4 |
| `SkewChildTranslateManipulatorTests.cs` | 2 |
| `MotionVariantPropagationTests.cs` | 2 |

#241 says ten. That number was mine and it was wrong: a grep for `Within` also matched the identifiers `GroupFocusWithin` and `PeerFocusWithin` in `StyleVariantClassTests`, which carry no tolerance at all. The analyzer finds every real site and no false one.

Getting that count right took three attempts, and the first two failures are worth stating because they were the same mistake in different clothes. A run against a warm `Library` reported **zero** sites, which reads as "the analyzer does not fire" and is really "nothing recompiled" — Unity emits diagnostics only for assemblies it actually builds. `touch`-ing a file per assembly did not fix it either, because Bee's cache is content-hashed and an mtime change invalidates nothing; that run reported 6, covering the single assembly whose content had genuinely changed. Only appending a line to one source per assembly produced a real rebuild of all sixteen and the full 8.

Verification that the analyzers reach the Unity assemblies at all is separate and direct: a deliberate 25-branch method in a marked test assembly yields `error VEL501`, alongside the VEL503 warnings, in the same compile.

Generator suite **396 passed / 0 failed** at head. The baseline is not quoted: the figure first put here was taken on a tree that was not the merge base, and a count that is not attributed to a commit is not a count. EditMode **3538 passed / 0 failed / 0 inconclusive**, no compile errors. `build.sh` run and the redeployed DLLs are committed.

Fixing the eight sites is #241 and is deliberately not in this change: an analyzer landing with its own violations already reported is what makes the backlog visible instead of assumed.


s4k10503:
A known gap for VEL503, found while fixing the sites it reports (#241).

The rule matches on the expected value's **type** being a `ValueTuple`. NUnit walks an expected **array** element-wise carrying the tolerance, so an array is not itself a trap — but when each element is a tuple, the comparison dies in exactly the same place: `new[] { ("a", 0.99999f) }` against `new[] { ("a", 1f) }` fails, where `new[] { 0.99999f }` against `new[] { 1f }` passes. The expected value's type there is an array, so VEL503 does not fire.

One such site exists in the package and is fixed on `fix/tuple-tolerance-assertions`. It was found by hand, not by the analyzer.

Not a request to widen the rule in this PR — a tuple *inside* a collection is a different match and worth its own decision, and the descriptor's help text already names what it does not catch. But that text should name this shape too, since it is the one shape a reader would reasonably assume is covered.

Issue #246, opened 2026-07-30
Every shader-backed paint silently does not render in a built player


s4k10503:
`Shader.Find` resolves all four Velvet shaders in editor Play Mode and returns **null** in a built player. Measured, not inferred.

The paints that depend on it therefore do not render in any build, with no error and no warning — the code takes its null path and draws nothing:

- drop shadows
- gradient silhouettes
- the shader-backed filters, `brightness-*` and `saturate-*`

A developer sees all three working in the editor for the entire life of a project and finds them missing the first time they ship. Nothing in the suite covers it, because every fixture runs in the editor, where `Shader.Find` succeeds.

This is the same class as #227, one layer down: an editor-only resolution path standing in for a runtime one. #227 was the stylesheet; this is the shaders. Both were invisible until something ran in an actual player.

## Why it happens

Unity's manual states it directly: "By default, Unity examines the scenes in the build and automatically strips shader variants that are not used in those scenes." A package shader reached only by name from C# is in no scene, so nothing keeps it. That is what *Always Included Shaders* exists for — but it is a project setting the consumer owns, so a package cannot ship it.

**A `Resources` folder is not the answer, and the first version of this issue was wrong to list it.** Unity documents the Resources system as something to avoid: it slows application startup and build times, makes memory management harder, prevents per-platform delivery and incremental updates, and its content is resident at runtime whether used or not. Addressables is the documented replacement — and `com.velvet.core` already depends on `com.unity.addressables` 2.7.4.

- <https://docs.unity3d.com/Manual/class-GraphicsSettings.html>
- <https://docs.unity3d.com/6000.5/Documentation/Manual/LoadingResourcesatRuntime.html>
- <https://docs.unity3d.com/2022.3/Documentation/Manual/UnderstandingPerformanceResourcesFolder.html>

What that leaves is a genuine design question nobody here has settled with evidence. Addressables asks the consumer to create groups and run an Addressables build, which a package cannot assume has happened; a serialized reference from an asset the consumer's scene already holds costs nothing but changes the public surface; a `ShaderVariantCollection` is a third shape with its own rules. Whichever is chosen, the reason has to come from a measurement in a built player, not from an argument.

Note that #239 shipped the stylesheet through a package `Resources` asset. It has the same problem and is #258 — do not copy it.

## Acceptance criteria

- Every shader-backed paint renders in a **built player**, shown by capture or pixel readback from the build — not from Play Mode, which is where this hid.
[...]

PR #254, opened 2026-07-31
fix(velvet): put the bundled shaders in a build without a Resources folder

s4k10503:
Addresses #246. **The first version of this branch moved the shaders into a `Resources` folder. That was wrong and is fully reverted here** — the four shaders are back at `Runtime/Styles/Shaders/` with their original GUIDs, and `Runtime/Resources/Velvet/` keeps only the stylesheet, which is #258's problem and not this one.

`Shader.Find` resolves all four Velvet shaders in editor Play Mode and returns null in a built player, because Unity strips shader variants no scene references and a package shader reached only by name from C# is in no scene. Drop shadows, gradient silhouettes and the `brightness-*` / `saturate-*` filters therefore take their null path and draw nothing in every shipped build, announced only by a warning in the player log and invisible in the editor.

## The mechanism, and what it costs the consumer

An `IPreprocessBuildWithReport` injects the four shaders into Graphics Settings' **Always Included Shaders** for the duration of the build; an `IPostprocessBuildWithReport` removes them again. `VelvetShaders` owns the four names and the three call sites take constants.

**The cost, stated because #246 asks for it:** every player build carries all four shaders and their variants whether or not the project uses a shadow, a gradient or a filter. That is the price of a name-resolved shader in a package — Addressables would move the cost to the consumer's group configuration and an Addressables build this package cannot assume has happened, and a serialized reference would move it into the public API. An entry the consumer already had is recorded as theirs and left alone, so the revert never removes something it did not add.

## Three things the measurements changed

1. **The injection reaches `GraphicsSettings.asset` on disk during the build.** A killed build left the file carrying all four GUIDs. So the revert has to save as well as edit, or the consumer keeps a diff — the exact outcome the whole shape exists to avoid.
2. **A build that dies before its post-process leaves that diff.** The record of what to undo therefore lives in `Library/VelvetBundledShaderInclusion.txt` rather than in editor state, and a `SessionState` marker separates a dead session from a mid-build domain reload — repairing during a live build would reintroduce the defect it repairs.
3. **A read-only `GraphicsSettings.asset` fails the build up front**, with a `BuildFailedException` naming the file. Not because the write would fail, but because the *revert* must be able to write, and a revert that cannot leaves the diff.

## Verified

EditMode 3543 total / 3540 passed / 0 failed / 0 inconclusive / 3 skipped — the three skips are pre-existing `RouterHooksTests` cases, untouched here. PlayMode 154 / 154. `assert-no-inconclusive.py` clean on both. `GraphicsSettings.asset` hashes identically before and after seven inject-and-revert cycles.

**RED per cut, one at a time, reverted in a `finally`:** removing `Inject()` from the preprocess reddens 5 of 7 cases. Every other cut reddens exactly its own case and nothing else — no-post-revert reddens 2, no-save-on-revert reddens only the byte-identity case, no-read-only-check reddens the read-only case, unguarded-repair reddens the still-running-session case, a dropped shader name reddens the tree-walk case, revert-everything reddens the consumer-entry case. No cut left a case green that should have moved.

The tree-walk case is #246's third acceptance criterion: a shader added under `Runtime/Styles/Shaders/` without being reached by this mechanism fails it.

## What is not verified, and it is the headline

**#246's first acceptance criterion is not met. There is no proof from inside a built player, neither positive nor negative.**

Seven `-testPlatform StandaloneOSX` runs. One reported; six died with the player booted, Metal up, assemblies loaded, then silent — `Test execution timed out. No activity received from the player in 600 seconds`.

| run | `BuildInclusion` assembly | outcome |
|---|---|---|
| 1 | no | **reported** — 119 cases, 104 failed |
| 2, 3 | yes | timeout, player idle |
| 4, 5, 6 | yes | timeout, player idle |
| 7 (injection disabled) | yes | timeout, player idle |

Ruled out by measurement rather than by argument: contention (two of the runs started at zero editors and zero players with a sampler recording peak 1 throughout), `-testFilter` (unfiltered runs fail identically), `AssetDatabase.SaveAssets()` in the preprocess (removing it changed nothing), the build callback itself (run 7 has the injection disabled and fails identically), and a callback throw (neither editor log carries one). What remains is the *presence of the new editor test assembly* or an environment state that went bad before run 2. That was not isolated and is not guessed at here; it is recorded on #246 with the next step.

Four of the six timeouts also ran on a tree carrying a stray `BundledShaderBuildInclusion.cs.keep` inside the package, which Unity imported as an asset. It is present in four failures and absent in two, and every run carrying the new assembly failed either way — **a defect in the record, not the cause.**

**Also unverified: whether Always Included Shaders is sufficient for these shaders in a player.** Everything the documentation says points that way; nothing here confirms it from inside a build.

**And a live constraint on ever proving it, now #273:** `RenderTexturePanelHost` builds its `PanelSettings` with no theme, so in the one player run that reported, 102 of the 104 failures were `ArgumentNullException: Value cannot be null. Parameter name: shader` and the 15 passing cases are exactly the ones that never render. No pixel-readback fixture can be green in a player until that is fixed, which is why the fixture here reads `Shader.Find` and nothing else.

## Riding along, named so it is not a surprise

The branch also removes an `[Unreleased]` CHANGELOG entry about mid-play shadow co-fade. It is defensible on its own terms and it is not this change.

`Documentation~/player-builds.md` is rewritten — it described the `Resources` mechanism. The `Documentation~/README.md` index row follows it, and `styling-filters.md` gains a one-line link rather than a second copy. `CHANGELOG.md`'s entry is rewritten for the same reason.

---

**Corrected after merging.** This description said the pre-change failure was silent, "with no error and no warning". That is false: `DropShadowBaker`, `GradientSilhouetteBaker` and `BuiltInFilterDefinitions` each log a warning on exactly the null-shader branch, through an unconditional `Debug.LogWarning`, and this pull request's own CHANGELOG entry said so. The failure was announced by a warning in the player log and invisible in the editor — a weaker claim than the one made, and the true one.

**Also settled after merging, on #246:** the player-side proof this description records as not obtained *was* obtained. Two builds differing only in whether the injection runs: with it, all four shaders resolve and the three paints draw — the drop shadow puts 46 distinct colours outside its card, the gradient silhouette 333, and `brightness-150` over `bg-green-600` reads 33/245/111 against the declared 22/163/74, which is 1.5× exactly. Without it, all four `Shader.Find` return NULL and every one of those readings collapses to a single colour. The reason seven earlier attempts failed was a windowed player being unable to advance frames while the Mac's display is asleep, which `-batchmode` avoids.


s4k10503:
Holding this. The defect it fixes is real — I reproduced it myself rather than relying on the report, in a StandaloneOSX test build:

```
Shader not found: Velvet/FilterBrightness
Shader not found: Velvet/FilterSaturate
```

Three `BuiltInFilterShaderPlaybackTests` cases fail in the player and pass in the editor. Unity's manual states the mechanism: "By default, Unity examines the scenes in the build and automatically strips shader variants that are not used in those scenes." A package shader reached only by name from C# is in no scene.

**The remedy is wrong, though.** Unity documents the Resources system as the thing to avoid — it slows startup and build times, makes memory management harder, blocks per-platform delivery and incremental updates, and its contents are resident whether used or not. Addressables is the documented replacement, and this package already depends on `com.unity.addressables` 2.7.4. Choosing the discouraged mechanism when the recommended one is already a dependency is not something to merge because the checks are green.

- <https://docs.unity3d.com/6000.5/Documentation/Manual/LoadingResourcesatRuntime.html>
- <https://docs.unity3d.com/2022.3/Documentation/Manual/UnderstandingPerformanceResourcesFolder.html>

What is genuinely unsettled is which mechanism a package can rely on when the consumer does nothing but install it. Addressables asks for groups and an Addressables build; a serialized reference needs an asset the consumer's scene already holds; Always Included Shaders is a project setting a package cannot ship. A survey of how Unity's own packages and OSS UPM packages solve this is running, and the answer decides this PR and #258 together — #239 shipped the stylesheet the same way and has the same problem.

Also note #259: `-testPlatform StandaloneOSX` does not compile on `main` at all, because a fixture I added drives editor-only preview types from an all-platforms assembly. That is what stood between this repository and noticing any of this.

s4k10503:
Review pass. Four findings survived adversarial verification; two were killed. Three are fixed here, and one correction to this description.

**A revert that could not finish deleted its own record, which made the leftover permanent.** `Revert()` skipped a recorded name whose `Shader.Find` returned null — the package renamed or dropped that shader since the build — and then deleted the record anyway. Nothing afterwards could clean the orphan: `Inject()` only adds names that are missing, so on the next build it reads the entry as the consumer's own and adopts it. That is exactly the "nothing lands in your diff" promise, inverted. `Revert()` now rewrites the record with what it could not remove and returns, so a later pass can finish.

Pinned by `Given_ARecordedNameThatNoLongerResolves_When_TheRevertRuns_Then_ItKeepsTheRecord`. RED measured: with the record-keeping removed, that case fails and the other seven pass.

**The writability check was inside `Inject()`, but `OnPreprocessBuild` is `Revert(); Inject();` and the revert writes too.** With a leftover record and a read-only settings file, the revert mutated and saved before the refusal that should have stopped it. Hoisted to the top of `OnPreprocessBuild`.

**The guides and CHANGELOG asserted as settled the one thing this PR says is unverified.** `styling-filters.md` said both shaders "reach a player build"; `player-builds.md` said "a build strips a shader that no scene reaches" as fact; the CHANGELOG said the three paints "now render in a built player". All three now say what was actually shown — that the names are in Always Included Shaders while the build runs, pinned by a test — and `player-builds.md` gains a paragraph naming what has **not** been observed and what a consumer should do if a paint draws nothing.

**Correction to this description.** It said the pre-change failure was silent, "with no error and no warning". That is false: `DropShadowBaker`, `GradientSilhouetteBaker` and `BuiltInFilterDefinitions` each log a warning on exactly the null-shader branch, through an unconditional `Debug.LogWarning`, and this PR's own CHANGELOG entry said so. The failure was announced by a warning in the player log and invisible in the editor, which is a weaker claim than the one made and the true one.

**Two findings were killed by verification**, recorded so they are not re-raised: that the four pixel-readback PlayMode cases are guaranteed to fail in a player — they use a pre-existing shared host, so that is #273's condition rather than four new instances, and they do discriminate in editor PlayMode; and that the `SessionState` guard rests on unpinned engine semantics — it uses the API for its documented contract.

**A note on how the first attempt at this measurement went wrong**, because it is the same failure the rest of this session has been about. The first A/B runs reported the fix breaking a pre-existing case. It did not. `ProjectSettings/GraphicsSettings.asset` was carrying four leftover entries from earlier runs of the fixture, so `Unreached()` was empty before any test started and every arm of the comparison was measured against a dirty instrument. The fixture leaked because the new case left a record `Revert()` deliberately refuses to act on. `TearDown` now deletes the record, and a full run leaves the asset unmodified — checked, not assumed.

Full EditMode on the corrected branch: 3550 total, 3546 passed, 0 failed, 0 inconclusive, 3 skipped.

s4k10503:
The CI failure is diagnosed and fixed, and the cause was a defect in this change rather than in the runner.

**`FileInfo.IsReadOnly` does not answer "can I write this".** On Unix it reports the permission triad matching the process's own uid and gid. In the game-ci container Unity runs as **root** over a checkout owned by **uid 1001**, so the *other* triad of a `0644` file is `r--` and the attribute reads read-only for a file the same process writes freely. Measured in a Linux container rather than inferred:

```
start          : mode=644:1001:1001 IsReadOnly=True open(Write)=OK
after set true : mode=444:1001:1001 IsReadOnly=True open(Write)=OK
after set false: mode=644:1001:1001 IsReadOnly=True open(Write)=OK
```

All six CI failures carried one byte-identical message and one stack, from that predicate, and four of them execute *before* the case that sets the attribute — so the condition arrived with the container and the read-only case never created it.

**The attribute is wrong in both directions, and the second direction is the dangerous one for a consumer:**

| process | file | `IsReadOnly` | open for write |
|---|---|---|---|
| root | uid 1001, `0644` | **true** | **succeeds** |
| root | uid 1001, `0444` | true | **succeeds** |
| non-owner | uid 1001, `0600` | **false** | **UnauthorizedAccessException** |
| owner | own, `0644` | false | succeeds |

The third row is a build that passes the guard and then has `Revert()` mutate and save a file it cannot write — the permanent diff this whole mechanism exists to prevent. So writability is now established by attempting it: opening the file for write and closing it changes nothing and answers the question actually being asked.

**Three more defects came out of the same reading, all fixed:**

The check's own comment said it belongs at the top of `OnPreprocessBuild` because "Revert runs first and writes too" — but `OnPostprocessBuild` and the editor-load repair both reach `Revert()`, which saves, with no check at all. `Revert()` now declines and keeps the record when it cannot write, rather than throwing: at editor load there is no build to fail, and keeping the record is what lets a later pass finish.

`Inject()` committed the injection with `ApplyModifiedProperties()` **before** writing the record of what to undo. A failed record write left entries applied with nothing able to remove them. Reversed: a record naming something that never landed is cleaned up on the next pass; entries with no record are not.

`RequireWritableSettings()` inside `Inject()` was unreachable — `OnPreprocessBuild` is its only caller and calls it two lines earlier. Removed.

**The read-only case was a false green in CI**: the file already read as read-only before its arrangement ran, so it passed for the environment's reason rather than the code's, and deleting its arrangement would not have changed the outcome. It now holds an exclusive handle — which does bind, measured — and folds the arranged precondition into the assertion, so where the file was already unwritable the case fails instead of proving nothing.

**Deliberately not fixed in the workflow.** game-ci exposes `runAsHostUser` and `chownFilesTo`, and setting either would have turned CI green while leaving this class refusing a real player build in any root-in-container setup.

Full EditMode on the corrected branch: 3551 total, 3548 passed, 0 failed, 0 inconclusive.

**Not measured, and stated rather than assumed:** whether Unity's own settings save writes through a temp file and rename, in which case *directory* permission is what matters and a file-level probe answers a narrower question; and the Windows behaviour of any of this, including whether the `ReadOnly` attribute makes an open-for-write throw there.

PR #261, opened 2026-07-31
test: throwaway probe, do not merge

s4k10503:
Throwaway. Opened only to produce a BEHIND pull request so the pre-merge guard can be shown to refuse one. Closing immediately.


s4k10503:
Throwaway. It existed to produce a branch genuinely behind main so the pre-merge guard could be shown to refuse one, which it does — and which the guard's first version could not, because it keyed on mergeStateStatus and GitHub reports CLEAN for this PR despite eight commits behind.

Issue #265, opened 2026-07-31
A documentation-only change runs no test suite, including the guards written for documentation


s4k10503:
`DocumentationDriftTests` scans `CLAUDE.md`, both `README.md` files, `Documentation~/` and `Generators~/README.md`. CI does not run it for a change to any of them.

- `test.yml` excludes `**.md` and `Documentation~/**`, and the root `README.md` and `CLAUDE.md` are outside its path set entirely.
- `generators.yml` excludes `**.md`, with two hand-added re-includes — `Documentation~/setup.md` and, as of #264, `Generators~/README.md`.

So a rename that breaks a markdown reference is caught by a fixture that CI never runs for the change that broke it. The guards exist; the trigger does not. Both re-includes were added after somebody noticed the gap for one specific file, which is the shape of a rule being maintained by memory.

## Why it stayed invisible

A documentation-only PR reports all checks green, because the only jobs that would fail are the ones that never start. That is indistinguishable from a passing run, and it is the same reading-absence-as-success failure this repository has hit repeatedly — a warm `Library` reporting zero analyzer warnings, an empty check list read as "still running", a watcher alive and silent because everything had finished.

## What to establish

- Which markdown files any fixture reads, derived from the fixtures rather than listed by hand. `DocumentationDriftTests` already enumerates its corpus; the path filter should follow the same source instead of restating it.
- Whether the answer is to re-include those paths in `test.yml`, or to run the drift fixtures in a separate cheap job with no Unity licence requirement. The second is worth pricing: the fixtures are EditMode and need the editor, so it may not be available.
- The cost either way. Adding `**.md` back to `test.yml` means a typo fix in a guide triggers a 21–25 minute Unity run.

## Acceptance criteria

- A PR touching only a markdown file that a fixture reads runs that fixture. Prove it by pushing a deliberate break in one and watching CI go red, not by reading the path filter.
- The path set is derived from what the fixtures actually scan, or a test fails when the two disagree. A third hand-added re-include is the failure this issue is about.

s4k10503:
`scripts/` has the same gap and should be named alongside the markdown files.

`test.yml` does not watch it. So a change to `scripts/assert-no-inconclusive.py` — the script CI itself runs after every Unity job to turn an Inconclusive into a failure — triggers no Unity suite. The one file whose breakage would silently disable the repository's own guard against silently-passing tests is outside the path set.

[...]

PR #269, opened 2026-07-31
test: throwaway probe, do not merge

s4k10503:
Deliberate break: one guide, one renamed hook reference, nothing else in the diff. Open only to record what CI does for it. Close without merging.

### DIFF
--- a/Assets/Editor/URPSetup.cs
+++ b/Assets/Editor/URPSetup.cs
@@ -67,3 +67,4 @@ namespace Velvet.EditorTools
     }
 }
 #endif
+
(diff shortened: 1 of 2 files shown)


s4k10503:
Closing. Throwaway, and confounded as constructed.

It changes two files: the stale identifier in `Documentation~/focus.md` that was the point, and a blank line appended to `Assets/Editor/URPSetup.cs`. `Assets/**` is already in `test.yml`'s path set and this branch carries the **old** filter, so the run started because of the `.cs` file. What it shows is that `DocumentationDriftTests` catches a stale identifier — never in doubt — not that a markdown-only change starts a run, which is the claim #265 exists to settle.

A replacement branched from `chore/ci-path-set-drift-guard` and touching **only** a markdown file answers it. A second touching only `scripts/assert-no-inconclusive.py` answers the other half. Neither needs the local machine.

PR #270, opened 2026-07-31
test(velvet): give each absent-thing assertion a term the live mechanism produces

### LINKED ISSUE #245: Fifteen assertions pass against the mechanism they exist to pin
Fifteen cases across three fixtures stay green when the feature under test is neutered. They are not `Assume`-gated — folding does not reach them — because each asserts a **negative**, and a dead mechanism produces exactly the negative they expect.

The counts only mean something with the neuter cut attached, so they are given that way. Enumerated by a harness that applies each cut, runs the fixtures and diffs the per-test results, not by reading:

| feature | applier cut | parser cut | union |
|---|---|---|---|
| clip (`ClipPathWrapTests`) | **5** | 3 | 5 |
| ring (`ClipPathWrapTests`) | **4** | 3 | 4 |
| ring (`StyleRingClassTests`) | n/a | **6** | 6 |

**Why the cut matters, and why single-cut sweeps undercount.** Every hole found across three rounds was cut-dependent in the same direction: green at the applier cut, red at the parser cut. The reason is structural — a gate read one layer *up* survives a neuter one layer *down*. `Given_ClipOnMotion` and `Given_ClipAndShadowClasses_..._NoShadowPaintIsAttached` both keep passing with the applier dead because the warning and the suppression flag they depend on are gated on `StyleClipPathClass.WantsClipWrapper`, the parser. This is not a property of these fixtures. It is what happens whenever a test's discriminating term is a side effect of a different layer than the one under test — an expected log line, a suppression flag — and it will keep hiding exactly the cases whose red comes from somewhere other than where the test is aimed.

**The count moved on every recount.** It was three, then four, then five for the clip half, and the ring parser fixture went five to six when `Given_PlainUtility_When_GateChecked_Then_NotARingClass` turned up — the same negative shape as its five `TryExtract` siblings, one method over. Two inflations were also corrected: matching candidacy against comment text pulled in a ring test whose comment mentions clip, and asking the applier cut of a parser-only fixture reported 18 phantom holes. A number taken by reading has been wrong every single time.

## Why folding does not reach them

The sentinel used for the rest of `StyleRingClassTests` structurally cannot help: it substitutes for the helper's return value, and these cases never call the helper. Routing them through it would only assert "the spec is the no-ring sentinel", which restates "the parser said no" and is equally satisfied by a dead parser.

## What closing them takes

Each needs a **positive control in its own assertion** — a term the working mechanism produces and the neutered one does not, compared alongside the negative. That is a fixture restructure per case, not a fold, which is why the `Assume` work deliberately stopped short of it.

## Acceptance criteria

- Every one of the fifteen fails under the cut named for its fixture. Report results per cut; a total without cuts is not checkable.
- The control term is something the feature produces, not something the fixture arranges — a control the test sets up itself holds whether or not the code works.
- No case loses its single assert; fold the control in as one comparison.
- Re-run the enumeration after the fix rather than reasoning that the list is now empty. The harness exists; the count has never survived a recount.

Distinct from #241, which is a tolerance that never applied rather than an assertion that cannot fail.

s4k10503:
Closes #245.

Fifteen cases asserted only that something was missing, which a layer producing nothing satisfies exactly. Each now folds in a term the working layer resolves — a wrapper around a clipped sibling, the physical child count the invisible-child predicate discounts, a `ContainsKey` on the binding table, or a class list the parser does resolve — keeping one assert per case.

**The control has to come from the layer the case is aimed at.** The Motion warnings and the shadow and ring suppression gates all read the **parsers**, so a case leaning on one of those stays green with the applier dead. That is why a count of vacuous cases means nothing without the cut beside it.

A sixteenth was found by review after the first fifteen were fixed, and it is the same mistake one level over: `Given_ClipAndShadowClasses_When_Reconciled_Then_NoShadowPaintIsAttached` took its positive term from the **clip** layer while the absent thing it asserts belongs to the **shadow** layer, so a wholly dead shadow create path satisfied it for a reason unrelated to the clip. It now mounts a shadowed sibling with no clip, and asserts exactly one surviving binding: zero says the shadow layer is dead, two says the clip did not suppress.

**Measured from the committed harness**, five cut-and-fixture pairs on this branch, machine quiet throughout:

| fixture | cut | in scope | still failing to notice |
|---|---|---|---|
| `ClipPathWrapTests` (39 pass at baseline) | clip-applier | 21 | **0** |
| | clip-parser | 21 | **0** |
| | ring-applier | 19 | **0** |
| | ring-parser | 19 | **0** |
| `StyleRingClassTests` (23 pass at baseline) | ring-parser | 23 | **0** |

**A false comment corrected, outside this diff but found reading it.** `SilhouetteBoundsSpacer.NonSpacerChildCount` carried "No production caller remains". `FiberZLayerCoordinator` calls it, and its own comment two hundred lines away explains why it wants exactly that physical reading — a z-managed element hosts its band and its own container. The same file also called it "the single centralized consumer", contradicting itself. It now names the caller and keeps the part that matters: anywhere else in reconciler code this gives the wrong number.

No behaviour change, so no guide update and no CHANGELOG entry.


s4k10503:
**The two fixes this PR shipped without a cut run are now measured, from the committed harness rather than a scratchpad.**

`scripts/neuter-check.py` on this branch with the harness merged in, five cut-and-fixture pairs, machine quiet throughout (peak other runs 0 on every run):

| fixture | cut | in scope | still passing |
|---|---|---|---|
| `ClipPathWrapTests` (39 pass at baseline) | clip-applier | 21 | **0** |
| | clip-parser | 21 | **0** |
| | ring-applier | 19 | **0** |
| | ring-parser | 19 | **0** |
| `StyleRingClassTests` (23 pass at baseline) | ring-parser | 23 | **0** |

Every case in scope for a cut goes red under it. The two fixes described here as "vacuous by reading, which this issue's own history says has been wrong every time" are both ring cases and both reddened.

**Two corrections to how that number was obtained, because the first two attempts produced different answers and the difference is the finding.**

The harness's first run reported **18 holes**, every one a ring case reported under the *clip-applier* cut. That is the phantom-hole trap this issue names, one level finer than the cut map expressed it: the map was per fixture, and `ClipPathWrapTests` holds cases for both mechanisms. Scope is now declared per case within a mixed fixture, and a guard fails when a case in such a fixture matches no scope.

Its second run reported **2 holes**, both in `StyleRingClassTests`, both calling `HasRingClass`. Not vacuous — the cut named itself "the ring class parser resolves nothing" while neutering only `TryExtract`, so the gate stayed live and the two cases that read it stayed green. The cut now takes `HasRingClass` as well, and one of those two cases is among the ones this PR fixed by hand, which is the loop closing: the incomplete cut was hiding exactly the case that had already been corrected.

Neither correction was reached by reading. Both came from the instrument disagreeing with a claim, which is the argument for landing it.

PR #274, opened 2026-07-31
test: throwaway probe, do not merge (markdown-only trigger)

s4k10503:
One guide, one renamed hook reference. Nothing else in the diff — the point is what CI does for a markdown-only pull request. Do not merge; close when it has been read.


s4k10503:
Purpose served. Closing.

This branch is a single-file diff touching only `Documentation~/focus.md`, with a deliberately stale identifier. It recorded **zero workflow runs** before #267 and, after it, started a Test run whose EditMode job went red for exactly the intended reason:

```
Documentation~/focus.md: UseFocusRingState
  Expected: <empty>
  But was:  < "Documentation~/focus.md: UseFocusRingState" >
```

Same branch, same diff. The only thing that changed is the path filter on `main`, so this is the positive half of #265 — a guard that reads the guides now runs when a guide changes, and catches the break when it does.

One thing learned constructing it: reopening a pull request does not re-evaluate the filter. A `synchronize` event is needed, so the proof required a second commit to the same file rather than a reopen.

PR #275, opened 2026-07-31
test: throwaway probe, do not merge (scripts-only trigger)

s4k10503:
One line added to the script both Unity jobs run as their last step. Nothing else in the diff. Do not merge; close when it has been read.

### DIFF
--- a/scripts/assert-no-inconclusive.py
+++ b/scripts/assert-no-inconclusive.py
@@ -87,3 +87,4 @@ def main(arguments):
 
 if __name__ == "__main__":
     sys.exit(main(sys.argv[1:]))
+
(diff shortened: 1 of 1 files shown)


s4k10503:
Purpose served. Closing.

A single-file diff touching only `scripts/assert-no-inconclusive.py`. **Zero workflow runs** before #267; a Test run after it.

That file is what both Unity jobs invoke as their final step to turn an Inconclusive result into a failure, and nothing else in the repository runs it. Before #267 a change breaking it merged with nothing having executed it, and the breakage would have surfaced later on an unrelated pull request as a failing step.

Still open and stated in #265: `scripts/**` starts the Unity suites, which catches a results script that *crashes* — but not one that wrongly *passes*, which is the failure that matters for this file. There is no test for it and no CI job runs Python.

PR #281, opened 2026-07-31
feat(velvet): ship a scene a developer can open and press Play on

### LINKED ISSUE #226: No player-side starting point ships with the package
## Correction to the original framing

An earlier version of this issue claimed the repository has no runnable surface. That is wrong: **Window ▸ Velvet ▸ Preview** mounts `[VelvetPreview]` stories on a live panel without entering Play Mode, and `Editor/Preview/Examples/PreviewExampleStories.cs` ships examples to open it on. What follows is what is actually missing.

## What is missing

The preview window is editor-only, and nothing covers the player side:

- `Assets/` contains no scene (`find Assets -name '*.unity'` returns nothing), so nothing in this repository is ever entered in Play Mode by hand.
- The package has no `Samples~` directory and no `samples` array in `package.json`, so Package Manager offers a user nothing to import.
- No document names something to open and press Play on.

A developer who installs `com.velvet.core@1.6.0` gets a working framework and no working starting point: they must assemble a `UIDocument`, a `PanelSettings`, a mount call and a stylesheet reference from separate documents before seeing one element. The stylesheet half of that is #227, which has no supported runtime path at all.

## Why it matters beyond onboarding

A story in the preview window exercises mount, hooks, variants and Motion — but not `UIDocument` hosting, not `PanelSettings`, not scene load, not a player build, and not anything `#if UNITY_EDITOR` excludes. Those are exactly the seams a user meets first, and no test or tool touches them today.

## Acceptance criteria

- `Packages/com.velvet.core/Samples~/` contains at least one importable sample, declared in `package.json`'s `samples` array with `displayName`, `description` and `path`, so it appears in Package Manager.
- The sample includes its own scene, `PanelSettings`, and a mounting `MonoBehaviour`; opening it and pressing Play renders without further setup, and it demonstrates the stylesheet path #227 settles.
- Its content exercises state, an effect, a store, a keyed list, a route, a pointer-reactive variant and a transition — not a static label.
- `Documentation~/README.md` points at it as the starting point, and the package `README.md` names it in the install section.
- A PlayMode test loads the sample's scene and asserts it mounts. Note that Unity does not import `~`-suffixed folders, so state how the test reaches the scene rather than assuming it can.

s4k10503:
Closes #226.

The package shipped no player-side starting point. `Documentation~/setup.md` describes the wiring, and the preview window renders stories in the editor, but nothing existed to open and press Play on — the first thing a developer wanting to know whether this works at all reaches for.

**Starter App** is a scene with a `UIDocument` host, a `PanelSettings` and a theme, mounting a two-route app over a store: a keyed list, local state, an effect, hover variants, an enter/exit transition, and the bundled utility stylesheet attached from code rather than dragged onto the asset. It is registered in `package.json` under `samples`, so Package Manager offers it as an import.

**The trap this had to solve is that `Samples~` is never compiled.** Unity does not import a `~`-suffixed folder, so a sample shipped only there is never built, never opened and never laid out by the project that ships it — it can rot into something that does not compile and nothing says so. So the sample lives twice: `Assets/VelvetStarterSample` is the copy this project compiles, plays and tests, and `Packages/com.velvet.core/Samples~/StarterApp` is the shipped mirror. `scripts/sync-starter-sample.sh` makes the second equal to the first, and `StarterSampleShippingTests` compares them **byte for byte** and fails when they diverge — the same rule as the generated style table, which is that an unpinned mirror drifts silently.

`StarterSampleShippingTests` also walks `package.json`'s `samples` array and fails when an entry has no display name, no description, or a `path` that is not on disk.

**What the PlayMode cases pin**, loading the real scene and entering play mode rather than building a panel by hand:

- the mounted tree reaches the `UIDocument`'s panel;
- the header resolves `flex-direction: row` from the bundled sheet — read from a USS-backed class and not an arbitrary value, because an arbitrary value lands as inline style and would pass against an entirely unstyled panel;
- clicking the About link swaps the route.

That third case originally asserted only that an About-route element had appeared, which an outlet that **appends** rather than replaces satisfies exactly, and which the nav — living in the layout, present on both routes — could not distinguish. It now compares three terms at once: the task route was up, it left, and the About route arrived. Arrival alone and departure alone are each satisfied by a broken outlet; the conjunction is not.

`README.md` points at the sample as the way to skip the wiring, and its repository-layout tree gains `Samples~/`. No `Documentation~` guide changes: the sample adds no API and `setup.md` already owns the wiring it demonstrates.


s4k10503:
The `source-generators` job was red on `CodeShapeOptInDriftTests`, which enumerates every asmdef under the package and requires each to carry the code-shape opt-in marker. The sample's asmdef did not.

**The marker cannot go on it.** It raises VEL500–503 to `DiagnosticSeverity.Error` in the assembly that carries it, and a sample is imported into the consumer's project as source they own and edit. Marking it turns this repository's house limits — nesting depth, branch count, parameter count — into compile errors in a project that never adopted them, from a rule whose origin they have no way to see. The two copies are compared byte for byte, so there is no spelling where the shipped one is unmarked and the local one is not: **the starter sample is not shape-checked, and that is the trade this makes.**

So the enumeration now excludes asmdefs belonging to a sample the manifest declares, derived from `package.json`'s `samples[].path` rather than from the directory name. A folder under `Samples~` that the manifest does not declare is not shipped as a sample and stays in scope.

That exclusion can be pointed at the wrong thing, and a wrong one drops real assemblies from the guard silently. `Given_TheExcludedAsmdefs_When_Enumerated_Then_NoneIsAnAssemblyUnityCompiles` rejects that by requiring every dropped asmdef to sit under a `~`-suffixed folder, which is Unity's own rule for what it does not compile — so it asks a question the manifest cannot answer for itself.

Measured rather than argued: with the manifest pointing its one sample at `Runtime`, both that test and the marker test fail. With a valid sample entry **plus** a second one naming `Runtime`, the marker test stays green — 7 passed — and only the new test fails. That second case is the one it uniquely carries. Full generator suite: 407 passed, 0 failed.

PR #283, opened 2026-07-31
chore(velvet): disable a mechanism and report which of its tests still pass

### LINKED ISSUE #277: The neuter-cut harness is rebuilt from scratch every time, and its anchors are pinned by nothing
Three sweeps for vacuous tests have each rebuilt the same instrument from scratch, in a scratchpad that dies with the session. It applies a feature-level neuter at a named layer — the applier, the parser — runs the fixtures, reverts in a `finally`, and diffs the per-test results. About three minutes for four cuts.

This is not the mutation harness in #276. That one mutates syntax across a diff; this one disables a named mechanism at a chosen layer, which is what answers "does this assertion still hold when the feature is dead". Both are needed and they find different things.

## Why it is worth landing rather than rebuilding

**Every count taken by reading has been wrong.** Three, then four, then five for the clip half; five, then six for the ring parser fixture. And every rebuild has walked into the same two traps independently:

- matching candidate fixtures against text that includes comments, so a ring test whose comment mentions `clip-path-*` counts as a clip test;
- asking a cut of a fixture that never reaches that layer, which reported 18 phantom holes in one rebuild and 22 in another.

Both are mechanical mistakes with mechanical fixes, and both were re-made because the fix lived in a scratchpad.

**The structural finding it produced is the reason to keep it.** Every vacuous test found across the three sweeps was cut-dependent in the same direction — green at the applier cut, red at the parser cut — because a gate read one layer up survives a neuter one layer down. A single-cut sweep therefore misses exactly the tests whose red comes from somewhere other than where they are aimed, which is why a count means nothing without the cut beside it.

## The real objection, and its shape

The harness hardcodes about a dozen exact source anchors with nothing pinning them. It fails loudly at apply time when one stops matching, so it does not go quietly wrong — but it fails at next-use rather than at CI time, which means a rename can sit broken until the next sweep.

That is the same problem `Generators~/src/Velvet.StyleTable` already solves for the stylesheets: a fact the source owns is re-derived rather than restated. The answer is a guard that re-derives each anchor and fails when one stops matching exactly once, not discarding the harness.

## Acceptance criteria

- The harness lands under `scripts/`, with the fixture-to-cuts mapping explicit so a parser-only fixture is never asked an applier cut.
- Every anchor is re-derived or pinned by a test that fails in CI when it stops matching exactly once. An anchor that matches twice is as broken as one that matches zero times.
- It waits for a quiet machine **before** applying a cut, not after — a long wait otherwise leaves a mutated runtime file in the tree and blocks committing.
- It records concurrency for the duration of each run rather than sampling once at launch, and reports the peak, so "the machine was quiet" is measured rather than claimed.
- Running it over the two fixtures already fixed in #270 reports zero at every cut, reproducing that result from the committed harness rather than from a scratchpad copy.

s4k10503:
Closes #277.

Three sweeps for vacuous tests rebuilt the same instrument from scratch in a scratchpad that dies with the session, and each rebuild independently walked into the same two traps. Both are mechanical mistakes whose fixes lived nowhere.

This is not `scripts/mutation-check.py`. That one mutates syntax across a diff and asks whether any test depends on one line. This one disables a named mechanism at a named layer and asks the question a fixture's *name* makes: with clip-path resolving nothing, does a case called "the element is wrapped" still pass? A mutation lands inside one method, so it cannot ask about a mechanism spread over two files; a cut cannot ask about one boundary condition.

## The reproduction #277 asked for

Run against the tree carrying #270's fixes, five cut-and-fixture pairs, machine quiet throughout (peak other runs 0 on every run):

| fixture | cut | in scope | still passing |
|---|---|---|---|
| `ClipPathWrapTests` (39 pass at baseline) | clip-applier | 21 | **0** |
| | clip-parser | 21 | **0** |
| | ring-applier | 19 | **0** |
| | ring-parser | 19 | **0** |
| `StyleRingClassTests` (23 pass at baseline) | ring-parser | 23 | **0** |

## Three defects this harness found in itself, which is the argument for landing it

Its **first** run reported **18 holes**, every one a ring case reported under the *clip-applier* cut. That is the phantom-hole trap #277 names, one level finer than the cut map expressed it: the map was per fixture, and `ClipPathWrapTests` holds cases for two mechanisms. Scope is now declared per case inside a mixed fixture, and `NeuterCutAnchorTests` fails when a case in such a fixture matches no scope. Matching *two* is not a failure — a case named for a ring suppressed by a clip belongs to both and is correctly asked both cuts.

Its **second** run reported **2 holes**, both calling `HasRingClass`. Not vacuous: the cut called itself "the ring class parser resolves nothing" while neutering only `TryExtract`, so the gate stayed live and the cases reading it stayed green. A cut now takes the gate its summary names. One of those two is a case #270 had already corrected by hand — the incomplete cut was hiding the very fix it should have confirmed.

Interrupting a sweep **left a neutered parser in the working tree**. The revert runs in a `finally`, which a killed process does not reach. A sweep now refuses to start when git reports any cut file already modified, naming it — otherwise the next run measures a mutation nobody declared and reports its survivors as holes.

A fourth, smaller: `IsRingClass` and `IsClipPathClass` are expression-bodied, so a neuter inserted after "the line following the anchor" would have landed outside the body. The brace guard rejected them before any editor started, and the cuts take the brace-bodied gates instead.

## What is pinned in CI

`NeuterCutAnchorTests` fails when an anchor stops matching **exactly once** — twice is as broken as zero, since the harness would neuter one of the two and report the other's tests as holes — when the line after an anchor is not a body brace, when a fixture names an undeclared cut, when a fixture name is absent from the test sources (because `-testFilter` selecting nothing runs no tests and reports no holes), and when a case in a mixed fixture matches no scope. Each folds its own count into the comparison, since a map that parsed to nothing satisfies every one of them.

## Two things the acceptance criteria asked for by name

The wait for a quiet machine happens **before** the cut is applied, never after: a wait with a neutered file already in the tree leaves the working copy mutated for its duration. And concurrency is sampled every three seconds for the duration of each run and reported as a peak rather than read once at launch — an editor that starts *after* this run does is exactly the one that would make its failures ambiguous.

The baseline must be green before any cut is applied. A test failing on its own accord is indistinguishable from one the cut killed and scores as coverage the fixture does not have, which hides holes rather than inventing them.

No `Documentation~` change: this is development tooling, not package behaviour. `CONTRIBUTING.md` gains it beside the mutation harness, and the test skill points at it so a fourth rebuild does not happen.


s4k10503:
Review pass. Seven findings survived verification; all seven are fixed here.

**The clip-parser cut neutered a function nothing calls.** `HasClipPathClass` has zero production callers — its only mentions outside its own declaration are two comments — so that edit changed nothing the reconciler runs, while `TryExtractLive`, which resolves the mask actually applied, stayed live under a cut whose summary said the parser resolved nothing. The cut now takes `TryExtractLive`, and **`HasClipPathClass` is deleted** along with the two cases that tested it and the two comments that named it as the gate. `TryExtract` stays in the cut and was never inert: `WantsClipPath` and `WantsClipWrapper` both reach it, and `FiberNodeFactory` reaches them.

**A case that asked nothing was scored as killed.** `report_pair` counted a hole only when the cut result was `Passed`, so a case that went Inconclusive, Skipped or Ignored under a cut was folded into "killed" — coverage the fixture does not have, in the direction that *hides* holes. The asymmetry was unreasoned: the baseline check three lines away treats every non-`Passed` result as disqualifying and says why. Only `Failed` is now evidence, and a hole that was not `Passed` prints its actual result beside it. `scripts/mutation-check.py` already took this position with its `survived (inconclusive)` verdict.

**The fixture-existence guard compared class names against file names.** That is not an invariant here — `StyleRingClassTests.cs` declares a second fixture, `RingGeometryPanelTests` — so it rejected legitimate fixtures and passed renamed ones. It now searches for the class declaration. Both directions measured: pointing the map at `RingGeometryPanelTests` (real, in a differently named file) now **passes** where the file-name guard would have failed, and pointing it at a name nothing declares fails. `CaseNames` rested on the same assumption and now shares the lookup.

**A cut whose mechanism has no scope silences its whole fixture.** `in_scope` filters scopes by mechanism, so a third mechanism added to a mixed fixture without a scope entry puts every case out of scope; the sweep prints "0 of 39 in scope", finds nothing and exits 0, which reads exactly like a clean result. A new guard fails when a mixed fixture asks a mechanism it scopes no case to — measured by adding exactly that cut, which reddens it.

**Two guards asserted a bare `Is.Empty`** while the PR body claimed all of them fold a count in. Both now fold one, so the sentence is true rather than the sentence being softened.

**A false universal, in two places.** "Every vacuous test found here so far was green at the applier cut and red at the parser cut" is wrong: `Given_NoClipClass_When_Reconciled_Then_ElementIsNotWrapped` and `Given_ClipPathNone_..._ElementIsNotWrapped` have no term any clip cut at any layer can move, so they are green at both. Corrected in `CONTRIBUTING.md` and in the script's own docstring to say that some are cut-dependent and others are green at every cut — which is still the argument for asking more than one, and is true.

**A comment gave a reason that cannot occur here.** The anchor on `UNITY_RUNNING` was justified as preventing a self-match; the waiting process here is `python3 scripts/neuter-check.py`, whose command line contains neither the editor path nor `-runTests`, and the editor this script launches is compensated for by arithmetic elsewhere. The sentence was carried over from the sibling harness, where it is true. Deleted rather than reworded.

Full EditMode after all of it: 3547 total, 3544 passed, 0 failed, 0 inconclusive.

PR #284, opened 2026-07-31
feat(velvet): give the semantic colours a background, opacity and two sets

### LINKED ISSUE #230: The semantic colour tokens cannot render correctly on their own
## Scope

This is about `Runtime/Styles/_tokens.uss`'s semantic layer (`--color-surface`, `--color-text`, `--color-border`, …), not about `_palette.uss`, which carries Tailwind v3.4's full opaque hue × shade scale and is not in question. Utilities drawn from that scale are unaffected by everything below.

Of the 31 `--color-*` declarations in `_tokens.uss`, 27 are `rgba` and 4 are `rgb`. The translucent ones are white overlays — `--color-surface: rgba(255, 255, 255, 0.12)`, `--color-border: rgba(255, 255, 255, 0.36)` — designed to composite over a darker backdrop. As a theme that is a coherent choice, and `_tokens.uss` already states that a consumer's own `:root` wins on later source order. Three consequences are not choices.

## 1. No backdrop token is declared, so the layer cannot render itself

There is no `--color-background`. Every translucent token presumes a surface underneath that nothing in the package supplies and no document mentions. Mounted on a bare panel, `bg-surface` composites 12%-white onto whatever is behind, and `text-strong` — an opaque near-white — lands on it.

This is observable: the repository's story capture renders the example stories, which use `text-strong` / `text-subtle`, and without an explicitly supplied dark backdrop they capture as very nearly blank while passing a check that asks whether any pixels differ. A user meets the same thing on their first mount, with no capture to show them (#227 is the stylesheet half of that first mount).

## 2. Stacking is not idempotent

Two nested elements both carrying `bg-surface` composite to roughly 0.226 alpha, not 0.12. Two components that are identical in isolation render differently by nesting depth, and a refactor that adds or removes a wrapper changes colour silently.

`bg-slate-800` has no such behaviour, and no C# or UI Toolkit constraint forces the deviation — which is the bar CLAUDE.md sets for departing from the web's semantics.

## 3. The layer is dark-only by construction

`--color-text` is `rgb(242, 247, 255)`. There is exactly one `:root` block in `_tokens.uss` and no light counterpart anywhere, and `dark:` is a variant manipulator that toggles *classes* — it does not swap a token set. So an application built on the semantic tokens has no path to a light theme short of replacing the block wholesale, and nothing says so.

Relatedly, a contrast ratio cannot be computed for a translucent token without a declared backdrop, while `_accessibility.uss` exists and implies the project cares about one.

## What the parity target does

`CLAUDE.md`'s first principle settles this rather than leaving it to taste: "reproduce React's semantics as faithfully as possible… When unsure whether a behavior is 'correct,' the answer is 'what React does.'" For utility-first styling the upstream is Tailwind, and for a semantic token layer over it, shadcn/ui. Both answer all three points, and they answer them the same way:

- **Opaque values, not translucent overlays.** Tailwind's palette is absolute colour; shadcn's semantic tokens (`--background`, `--foreground`, `--card`, `--border`, `--primary`, …) are opaque. Nesting is therefore idempotent — `bg-card` inside `bg-card` is the same colour, where `bg-surface` inside `bg-surface` here composites to roughly 0.226 alpha.
- **A declared background.** shadcn declares `--background`. Velvet declares none, which is why the layer cannot render itself and why a capture of it reads as blank.
- **Two sets, selected by a class.** shadcn ships a light set and a dark set on `:root` and `.dark`. Velvet ships one set that is dark by construction — `--color-text` is an opaque near-white — and `dark:` here toggles *classes* rather than swapping tokens, so an application on the semantic layer has no path to light.

So this is a parity gap in three places, not a design question, and the direction is fixed.

## Acceptance criteria

- `_tokens.uss` declares a background token, and the semantic colours are opaque, so that nesting two elements carrying the same background utility produces the same colour as one. Pin that with a test — it is the property translucency silently breaks.
- A light set and a dark set exist, and the mechanism that selects between them is stated in `Documentation~/styling-variants.md`. If `dark:` cannot do it as it stands, say what it would take rather than declaring the gap acceptable.
- `_palette.uss` is out of scope: Tailwind's hue × shade scale is already parity-correct and untouched.
- The example stories render the same before and after on a dark backdrop, so the change is visibly a widening rather than a re-theme. The story capture writes PNGs for exactly this comparison.
- Contrast becomes computable once the tokens are opaque and a background is declared. `_accessibility.uss` exists; say whether anything now checks a ratio, or that nothing does.

s4k10503:
Closes #230.

Every semantic colour token was a translucent white or a translucent accent over nothing. There is no `--color-background`, so what those tokens composited against was whatever the host panel happened to sit on — and a surface inside a surface came out brighter than one on its own, because two 12% whites are not one 12% white. `bg-surface` did not name a colour; it named a lightening step.

**The palette was never the problem.** `_palette.uss` carries Tailwind v3.4's full opaque scale. The translucency was confined to the semantic layer in `_tokens.uss`, which is the layer a consumer actually spells.

## What the tokens are now

A **role** token — background, surface, text, border, primary, danger, success — is opaque, so an element carrying `bg-surface` inside another one resolves to the same colour rather than compositing to a brighter one. `--color-white-*`, `--color-overlay*` and `--color-shadow` keep their alpha: they name a colour and a strength rather than a role, and compositing over what is behind them is what a scrim and a shadow are *for*. Those are declared once for both themes.

There are two sets: light on `:root`, dark on `.dark`. Each dark value is the old translucent overlay resolved against the new `--color-background`.

**That backdrop is new in this change, so this is a visible change for every project rather than a restatement of the existing look.** The old tokens composited against whatever the host panel happened to sit on, which is the defect above; the dark values match only a project that already sat on slate-900, and only for a single unnested layer. On a black host `bg-surface` was `rgb(31,31,31)` and is now `rgb(44,51,68)`.

`VelvetStyleUtilities.BindThemeTo` keeps a root carrying the `dark` class exactly while `VelvetTheme.IsDark` holds. `AttachTo` calls it; a project reaching the sheet from a scene reference calls it itself. `DarkThemeClass` is public so an application setting the class from UXML or a scene names the same one the sheet declares.

## What the tests pin

- the dark set names nothing the light set omits — a token declared in one and not the other silently falls back to the light value under `.dark`, which reads as a theme bug in one place and correct everywhere else;
- no theme-varying colour is translucent, which is the defect itself;
- every text-on-background pair in each set clears AA contrast, measured from the declared values rather than from a rendered pixel;
- an element carrying the background utility resolves a darker opaque colour when dark mode turns on, and the root stops carrying the class when the flag turns back off;
- **a surface inside a surface lands one colour**, in both sets, on a live PlayMode panel. That is the idempotence the whole change exists for, and it is the case a token table alone cannot answer.

`Documentation~/styling-variants.md` gains the two-set model and the binding call; `setup.md` and `preview-tooling.md` follow the same call, and `CHANGELOG.md` records the colour change as behavioural for anyone who had built against the translucent values.

Verification is CI's EditMode and PlayMode on this branch.

---

**Corrected after merging.** An earlier version of this description said the dark set was "what the framework already looked like, now stated rather than accumulated". That is false for every host but one, and this description's own first paragraph says why: the backdrop those values are composited against did not exist before this change. The CHANGELOG entry that ships with it records the change as behavioural; the description should have said the same.

A second sentence, on the set-coverage test, is also withdrawn. It said a token declared in one set and not the other "silently falls back to the light value under `.dark`". The test checks names declared only in `.dark`, which under `:root` are simply undeclared — there is no light value to fall back to. Read the other way round the sentence is true of the mechanism, and it would condemn this change's own design, since `--color-white-*`, `--color-shadow` and the two `--color-overlay*` are deliberately declared once for both themes.


s4k10503:
Review pass. Four findings survived verification; all four are addressed.

**A folded count that could not fail.** The contrast case asserted `(measured.Count, allPairsClear)` against `(sets.Length * TextTokens.Length, true)` — the same expression on both sides, true for any content of either array. With `TextTokens` empty it reads 0 against 2 × 0, `All()` is vacuously true, and the case passes having measured nothing. It looked like the folded precondition this repository's convention asks for, and its two siblings genuinely are: they fold `dark.ContainsKey("--color-surface")`, which goes false when the parse returns nothing.

The count is now the literal `4`. Measured both ways: green as it stands, and emptying `TextTokens` now fails exactly that case where before it passed.

**A count in the CHANGELOG that was right attached to a description that was wrong.** "27 of its 31 `--color-*` values as translucent white overlays" — 27 of 31 do carry an alpha, but only twelve are white. Twelve are translucent accents (five `--color-primary-*`, three `--color-danger-*`, three `--color-success-*`, plus the near-white `--color-text-subtle`) and three are near-black. A consumer sizing the migration was told their `--color-primary` and `--color-danger` were white overlays. Counted on `main` rather than taken from the finding: 31 declarations, 27 with alpha, 12 white, 12 accent, 3 near-black. The entry now says that.

**Two claims in this description are withdrawn.**

*"The dark theme is what the framework already looked like, now stated rather than accumulated."* False for every host but one. Each dark value is the old overlay composited over `rgb(15,23,42)` — a backdrop **this PR introduces**, since the first paragraph of this very description says the old tokens composited against whatever the host panel happened to sit on. So the equality holds only for a project already sitting on slate-900, and only for a single unnested layer. On a black host `bg-surface` was `rgb(31,31,31)`; it is now `rgb(44,51,68)`. This is a visible change for everyone, which is what the CHANGELOG entry says and what the description should have said.

*The rationale given for the set-coverage test.* It said a token declared in one set and not the other "silently falls back to the light value under `.dark`". The test checks names declared only in `.dark`, which under `:root` are simply undeclared — there is no light value to fall back to. Read the other way round the sentence is true of the mechanism, and it would condemn this PR's own design: `--color-white-*`, `--color-shadow` and the two `--color-overlay*` are deliberately declared once for both themes. The test checks the direction that is a defect; the direction the sentence described is the design.

`SemanticTokenSetTests` and `DocumentationDriftTests` green after the changes.

PR #285, opened 2026-07-31
refactor(velvet): delete the child flattener nothing calls and pin the one that runs

### LINKED ISSUE #266: Nothing fails when the reconciler stops dropping null children
`ChildReconciler.FlattenAndFilter` decides whether a child array needs processing with

```csharp
if (n is null or FragmentNode) { needsProcessing = true; break; }
```

Drop `null` from that pattern and an array holding a null but no `FragmentNode` takes the early return a few lines below and comes back **with the null still in it**. The whole EditMode suite stays green through that change.

`cond ? node : null` is the idiomatic "render nothing" in this framework — `CLAUDE.md` states it as such, and `FlattenAndFilter` dropping nulls is what makes it work. So the behaviour the architecture notes single out has no test that fails when it stops happening.

## Provenance, stated because it decides what to trust

The mutation and the green result are from the agent measuring per-mutant cost for #243, on its tree. I confirmed the code path by reading it — the early return is real and a null does survive it — but **I did not independently re-run the suite against the mutation**, because the machine had four Unity runs in flight. Reproduce it before building on it.

## Why it is worth its own issue rather than a line in #243

It is not a mutant that harness would have generated. Its operator set is spaced binary operators, boolean literals and void-call deletion; pattern arms are not in it. So this hole was found by hand and would have survived the instrument built to find holes — which is the more useful half of the finding, and the reason for filing it where it will not be read as that tool's catch.

## Acceptance criteria

- A test that fails when `null` is dropped from that pattern. Prove it red with the mutation applied and green without.
- Check the neighbouring case while there: an array containing a `FragmentNode` but no null, and one containing both, since the same early return governs all three.
- Say whether a null child surviving into the reconcile actually breaks anything downstream, or is merely tolerated. If it is tolerated, the test should pin the filtering itself rather than a symptom, and the reason belongs in the fixture.
- Consider whether the mutation operator set in #243's harness should grow a pattern-arm operator. Not required here, but this is the evidence for it.

s4k10503:
Closes #266.

#266 reported that dropping `null` from `ChildReconciler.FlattenAndFilter`'s

```csharp
if (n is null or FragmentNode) { needsProcessing = true; break; }
```

leaves the whole EditMode suite green, and asked for a test that fails when it stops happening.

**The suite was right and the issue's premise was wrong. `FlattenAndFilter` has no production caller.** Its only call site is its own recursive helper. Everything else that names it is a comment. So the mutation changed nothing any user could observe, and a suite that stayed green was reporting the truth.

The first version of this pull request added a test that called `FlattenAndFilter` directly and argued at length that a mounted tree could not pin the behaviour. It pinned dead code, and the argument was backwards.

## Where the drop actually happens

`GeneralPathReconciler.ExpandInlineRecursive`:

```csharp
switch (node)
{
    case null:
        continue;
    case FragmentNode fragment:
        ...
```

That is the live "render nothing", and a mounted tree pins it exactly — it is the consumer, so removing it cannot be absorbed by one.

## What this does

**Deletes `FlattenAndFilter` and `FlattenAndFilterRecursive`.** The buffer pool's node-list region stays: `GeneralPathReconciler` rents from it.

**Adds three cases that mount and read the rendered children by name**, not by count — a count alone passes against an implementation that rendered the null as an element and dropped one that belongs. A null at the top level, a fragment with no null, and a null nested inside a fragment, because the recursive descent has its own null arm and a fixture that only ever put a null at the top level would leave it deletable with nothing going red.

**Corrects three places that named the dead method as the mechanism**: `CLAUDE.md`'s architecture note, `V.ListFragment`'s doc comment, and two comments in `ReconcileState`. A `#region For FlattenAndFilter` in `ReconcilerBufferPool` is renamed for what it now serves.

## Measured

| | total | passed | failed |
|---|---|---|---|
| the three new cases, unmutated | 3 | 3 | 0 |
| with `case null: continue;` removed from the live path | 3 | 1 | **2** |

The two that fail are the two carrying a null; the fragment-only case stays green, which is right — it has no null in it. Full EditMode after the deletion: 3545 total, 3542 passed, 0 failed, 0 inconclusive, 3 skipped (the pre-existing `RouterHooksTests` skips). `assert-no-inconclusive.py` clean.

## A guard that was silent, and why

`DocumentationDriftTests` is supposed to catch a `CLAUDE.md` naming something that no longer exists. With `FlattenAndFilter` deleted from `ChildReconciler` but `CLAUDE.md` still naming it, **it passed** — because `#region For FlattenAndFilter` still carried the token, and the guard resolves a backticked identifier by asking whether it appears in *any source file*, not whether it is a live declaration. With that region renamed, the same experiment fails `Given_ProjectMarkdown_When_ScannedForBacktickedIdentifiers_Then_EveryIdentifierAppearsInASource` as it should.

So the guard does hold `CLAUDE.md` — and a comment or a region name left behind by an incomplete deletion silences it. That is worth knowing before the next deletion, and it is not fixed here.

## Withdrawn from the earlier description

That this hole would have survived `scripts/mutation-check.py` because a pattern arm is not in its operator set. The claim is moot now — the mutation was of dead code — and it was also wrong: the harness's boolean-literal operator reaches the same predicate.


s4k10503:
**RED and GREEN, measured on this branch rather than taken on trust.**

The mutation is #266's, applied to `ChildReconciler.FlattenAndFilter`:

```
- if (n is null or FragmentNode) { needsProcessing = true; break; }
+ if (n is FragmentNode) { needsProcessing = true; break; }
```

`-testFilter Velvet.Tests.ChildFlattenFilterTests`, EditMode, machine quiet, reverted afterwards and the tree confirmed clean:

| | total | passed | failed |
|---|---|---|---|
| unmutated | 3 | 3 | 0 |
| mutated | 3 | 2 | **1** |

The one that fails is `Given_AChildArrayHoldingANullAndNoFragment_When_Flattened_Then_TheNullIsGone`. The other two stay green under the mutation, which is what the PR body predicted and why they are described there as controls rather than as regression cases: with a fragment in the array, processing happens regardless of the null arm, and the recursive pass filters the null on the way through.

#266 asked for this to be reproduced before anything was built on it, because the original green result came from another tree. It reproduces.

s4k10503:
Review pass. Three findings, all of which the rewrite above had already reached — recording them so the overlap is visible rather than looking like they were ignored.

**The fixture's remarks and this description justified reading the returned array with a mechanism that does not exist.** That is the finding this PR now *is*: `FlattenAndFilter` has no production caller, the drop happens in `GeneralPathReconciler.ExpandInlineRecursive`, and the dead code is deleted rather than pinned.

**The claim that this hole would have survived `scripts/mutation-check.py` because a pattern arm is not in its operator set.** Withdrawn. It is moot — the mutation was of dead code — and it was also wrong: the harness's boolean-literal operator reaches the same predicate.

**The fragment case arranged four distinguishable elements and then asserted only a count and an absence.** It now compares the rendered names in order, so an implementation that inlined the fragment's children in the wrong place, or put them outside their siblings, fails it. The same change went into the other two cases for the same reason: a count alone passes against an implementation that rendered the null as an element and dropped one that belongs.

Issue #250, opened 2026-07-31
CI does not gate merging, so nothing but attention keeps a red branch out of main


s4k10503:
`protect-main` carries `pull_request`, `non_fast_forward` and `deletion`. It has **no required status checks**, and the repository has `allow_auto_merge: false`. So:

- A PR can be merged while its checks are failing, or before they have run. Nothing on the server side objects.
- `gh pr merge --auto` is unavailable, and enabling it as-is would be worse than useless: with no required checks, auto-merge merges immediately rather than waiting ([cli/cli#8514](https://github.com/cli/cli/issues/8514)).
- Every mechanism that currently keeps a green branch from sitting unmerged, or a red one from being merged, is a local watcher and a Stop hook — attention, dressed up. Eight green PRs sat unmerged for nine hours behind exactly that arrangement.

The fix is to move the gate to where the fact lives. With required checks plus auto-merge, `gh pr merge --squash --auto` hands the PR to GitHub, which merges it when the checks pass and never merges it when they do not. The "green PR nobody merged" state stops existing rather than being guarded against, and a failing check cannot be merged past by anyone, including by a mistake.

## The trap that has to be solved first

**Do not simply mark the existing jobs required.** CI here is path-filtered on purpose — `generators.yml` runs for `Generators~/**` and `Runtime/**`, `test.yml` for package and project changes, and neither runs for a docs-only change. A required check that is skipped by a path filter stays `Pending` forever and blocks the PR permanently; GitHub's own documentation says not to use path filtering on a workflow that is required. PR #244 is a live example — it is docs-only, reports zero checks, and would be unmergeable under a naive configuration.

The standard answer is one aggregate job that is always required and always runs:

- a `required-checks` job in a workflow with no path filter, `needs:` every real job, `if: always()`, whose body fails unless each dependency's result is `success` **or** `skipped`;
- that single job name is the required status check;
- the real jobs keep their path filters and stay unrequired.

That also decouples the required-check name from the job names, so a workflow can be split or renamed without touching branch protection.

## Acceptance criteria

- A PR with a failing check cannot be merged, shown by attempting it.
- A docs-only PR with no applicable checks can still be merged, shown by attempting it — this is the case that a naive configuration breaks, so it is the one that must be demonstrated rather than assumed.
- `allow_auto_merge` enabled and `gh pr merge --squash --auto` verified end to end: the PR sits, the checks pass, GitHub merges it with the squash subject and body given.
[...]

PR #290, opened 2026-07-31
probe: demonstrate that a skipped dependency still satisfies the aggregate

s4k10503:
Throwaway. Forces the licence check to report false so unity-tests is skipped, which is the state a fork with no secret is in. Shows Required checks (Unity) passing anyway. Deleted once recorded.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,8 +50,9 @@ jobs:
       - id: check
         env:
           UNITY_LICENSE: ${{ secrets.UNITY_LICENSE }}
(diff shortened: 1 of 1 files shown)


s4k10503:
Recorded, and deliberately not merged — it carries a workflow edit that must never reach main. With the licence check forced to report false, the run shows 'Unity tests (${{ matrix.testMode }}): skipped' and 'Required checks (Unity): success'. That is the state a fork with no UNITY_LICENSE secret is in, and #250's fourth acceptance criterion.

PR #321, opened 2026-08-01
test(velvet): hold the analyzer solution to the rules it implements

### LINKED ISSUE #313: Generators~/src violates the code-shape rules its own analyzers enforce
VEL500 (nesting deeper than four) and VEL501 (more than twenty branching decisions) are opt-in per assembly, and the package's own runtime assemblies opt in. `Generators~` does not, and it would not pass:

- `BuildCandidate` — 27 branching decisions
- `IsImmutableCapture` — 23 branching decisions

Measured when the analyzers were written; not re-measured since, so the numbers may have moved.

The rules exist because this repository decided a member past those shapes is harder to review than it is worth. The assembly that implements them is exempt from them, which is the one place the argument is weakest.

## The order this has to happen in

The rule recorded when the limits were introduced: do not leave them as warnings. An assembly opts in at error severity, so the backlog has to reach zero before the marker goes in — otherwise the marker is added, the build breaks, and the marker comes back out.

So this is two pieces of work: bring those members under the limits, then add
`[assembly: System.Reflection.AssemblyMetadata("Velvet.CodeShape", "enforce")]` to the generator projects.

s4k10503:
Closes #313.

VEL500 (nesting deeper than four), VEL501 (more than twenty branching decisions) and VEL502 (more than six parameters) are opt-in per assembly. The package's runtime and every one of its test assemblies opt in. The projects that *implement* those rules did not — the one place the argument for them is weakest.

## Why this is a marker and not a restructure

The self-application wiring already exists and is documented in its own project file. `Velvet.SourceGenerators.Bootstrap` recompiles the analyzer sources under a second assembly name precisely so the declaring project can load them, because a project referencing itself as an analyzer is rejected by MSBuild with MSB4006 before any compile starts. All three projects — `Velvet.SourceGenerators`, `Velvet.SourceGenerators.CodeFixes`, `Velvet.StyleTable` — already reference it as `OutputItemType="Analyzer"`.

Only the opt-in marker was missing.

## The backlog is empty

The rule for introducing these limits is that an assembly opts in at error severity, so the backlog reaches zero before the marker goes in. Two members were on record as over the branch limit; both have since been brought under it. With the marker in place the whole solution builds clean.

## That the analyzers run is measured, not inferred

A clean build proves nothing on its own — an incremental build skips the compile and the analyzers with it. A member making twenty-five branching decisions was added and reported:

```
error VEL501: Member 'Probe' makes 25 branching decisions; the limit is 20
```

then removed. The generator suite passes with the marker committed.

### DIFF
--- a/Packages/com.velvet.core/Generators~/src/Velvet.SourceGenerators.CodeFixes/CodeShapeOptIn.cs
+++ b/Packages/com.velvet.core/Generators~/src/Velvet.SourceGenerators.CodeFixes/CodeShapeOptIn.cs
@@ -1,0 +1,5 @@
+// The rules this solution implements apply to it too. The wiring that makes that possible is
+// Velvet.SourceGenerators.Bootstrap, which every project here already references as an analyzer;
+// this marker is what turns the rules on, and it goes in last because they are error severity — an
+// assembly that opts in with a backlog cannot build.
+[assembly: System.Reflection.AssemblyMetadata("Velvet.CodeShape", "enforce")]
--- a/Packages/com.velvet.core/Generators~/src/Velvet.SourceGenerators/CodeShapeOptIn.cs
+++ b/Packages/com.velvet.core/Generators~/src/Velvet.SourceGenerators/CodeShapeOptIn.cs
@@ -1,0 +1,5 @@
+// The rules this solution implements apply to it too. The wiring that makes that possible is
+// Velvet.SourceGenerators.Bootstrap, which every project here already references as an analyzer;
+// this marker is what turns the rules on, and it goes in last because they are error severity — an
+// assembly that opts in with a backlog cannot build.
+[assembly: System.Reflection.AssemblyMetadata("Velvet.CodeShape", "enforce")]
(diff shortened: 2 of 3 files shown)


s4k10503:
Closing without merging. The change does nothing it claims to do, and the one thing it does change is a regression.

All three projects already opt in. `src/Velvet.SourceGenerators/Velvet.SourceGenerators.csproj:34`, `src/Velvet.SourceGenerators.CodeFixes/...csproj:31` and `src/Velvet.StyleTable/...csproj:22` each carry `<AssemblyMetadata Include="Velvet.CodeShape" Value="enforce" />`, which the SDK emits as exactly the assembly attribute this PR added by hand. `Generators~/README.md` documents the mechanism as two MSBuild items and gives the XML.

So "the projects that implement them did not opt in" and "only the marker was missing" are both false, and the evidence offered for the change — a planted 25-branch member reporting VEL501 — reports identically on the tree without it. The probe was real and was measuring work that had already landed.

Two things the duplicate would have cost:

- **The project-file switch dies silently.** With the marker in two places, the analyzers' gate returns true on whichever it finds first, so setting `Value="off"` in the csproj no longer turns the rules off. Verified: with `off` set and a violation planted, VEL500, VEL501 and VEL502 all still fire.
- **The bootstrap gets opted in against a documented exemption.** `Velvet.SourceGenerators.Bootstrap` compiles its sibling's `**/*.cs` through a glob, so the new file lands in it too. The README says it "is not itself opted in; its content is measured by the sibling's own compile". Inert today because the bootstrap loads no analyzer, and live the moment it gains one or the pair moves into `Directory.Build.props`.

The comment was false in three places besides.

#313 is answered by the tree as it stands, not by this. Closing that too, with the measurement.

Issue #324, opened 2026-08-01
A neuter sweep leaves the project settings its cut stopped anything from reverting


s4k10503:
Registering the build-inclusion steps as cuts and sweeping them left two tracked files modified after the harness had reverted every source file it touched:

```
 M ProjectSettings/GraphicsSettings.asset    +4 always-included shaders
 M ProjectSettings/ProjectSettings.asset     +1 preloaded asset
```

That is the cut working. `stylesheet-revert` and `shader-revert` neuter the method that takes the entries back out, the fixtures then inject through the real code path, and nothing puts them back. `neuter-check.py` restores the source it edited — `revert()` writes back the text it captured in `apply_cut` — and has no view of state a fixture wrote outside the source tree.

## Why it matters beyond the tidy-up

The residue is silent and it is in tracked files. A sweep followed by a commit picks it up, and `refuse-blind-git-add.py` only refuses the blind form — a targeted `git add ProjectSettings/…` would not be stopped. It also poisons the next measurement: an always-included shader list that already holds the entries makes an injection test pass without injecting.

`dirty_cut_files` exists for precisely this shape one step earlier — it refuses to start a sweep on top of a source file some earlier interrupted sweep left neutered. The same argument applies to what a sweep leaves behind, and nothing makes it.

## Shape of a fix

The harness knows which files it edited. It does not know which files a fixture will write, and enumerating them per cut is the maintained-by-hand list this repository keeps replacing with derived ones. What it can do without any such list is compare `git status --porcelain` before and after each cut and report — or restore — anything that changed and is not a file it edited itself. That covers project settings, generated assets and anything else a fixture reaches, without naming any of them.

Restoring rather than only reporting is the safer default here, since the harness already owns a revert step and the residue is by definition state a fixture did not clean up.

s4k10503:
A second sweep, with the repair and gate cuts added, left no residue. That does not narrow the problem — it widens it.

The difference between the two runs is which cut ran last. The first ended on `shader-revert`, so the fixture teardown — which calls `Revert()` — was still neutered when the last case finished, and the entries stayed. The second ended on `shader-repair`, leaving `Revert()` intact for the teardown, which cleaned up.

So whether a sweep leaves the project mutated depends on the order the cuts happen to be declared in. A reader who runs a sweep and sees a clean `git status` learns nothing about whether the next one will be clean.

PR #325, opened 2026-08-01
Register the build and balance mechanisms as neuter cuts, and close what the sweep found

s4k10503:
Four cuts were registered in `scripts/neuter-cuts.json`, all for clip-path and ring. The build-inclusion steps and the text-balance frame had none, so `neuter-check.py` — which asks whether a fixture notices the mechanism it is named for dying — had never been pointed at either.

## What was registered

Eleven cuts across five layers: injection, revert and the domain-reload repair for each build-inclusion twin, the writability gate they share, the reconciler`s attach-and-detach of the balance manipulator, and the manipulator`s own subscription and apply.

The layers are what the first pass got wrong, for the third time on this instrument. Mapping the EditMode balance fixture to the manipulator`s internals reported all nine of its cases as holes — true and about nothing, because that fixture pins attach and detach and the reconciler is what does those. The unwritable-settings case is the same story one layer up: the gate refuses before injection runs, so it takes a cut of its own and a case scope that keeps it out of the other three.

## What the sweep found

Five cases were green with the code they are named for disabled.

Both twins carry a case asserting that a record naming something unresolvable survives the revert. A revert that never ran leaves the same survival, so neither case could tell the fix from the mechanism being dead — and both were written to pin exactly that fix. Each now folds in the removal the same revert did make.

The three balance cases have the same shape one layer over: a co-present width utility survives text-balance`s teardown, and a reconciler that attached no manipulator has no teardown to survive. Each folds in the attachment, which is this repository`s own rule for a precondition that gates the behaviour under test.

Holes under the mechanisms these fixtures name:

| fixture | before | after |
| --- | --- | --- |
| `TextBalanceParityTests` | 9 | 2 |
| `BundledStyleSheetInclusionTests` | 7 | 3 |
| `BundledShaderInclusionTests` | 6 | 1 |

What remains is what a cut cannot reach: negative assertions no cut can falsify, a case whose subject is the pool reset, and cases that are statements about a sibling cut`s layer.

## Verification

Every number above is a `neuter-check.py` run, not a hand-broken line. Each fixture`s baseline was green before each cut, and the cases named as fixed moved from the hole list to the failing list under the cut that names their mechanism.

Two things the sweep surfaced are filed rather than fixed here: #324, a sweep leaving project settings a neutered revert never took back out, and the remaining negative-assertion cases, which need a different instrument than a cut.

No `Documentation~` impact: `CONTRIBUTING.md` already owns what the harness is and how a cut is scoped, and this adds cuts rather than behaviour.

### DIFF
--- a/.claude/hooks/block-stop-on-open-backlog.sh
+++ b/.claude/hooks/block-stop-on-open-backlog.sh
@@ -2,17 +2,16 @@
 # Blocks Stop while assigned work is open and nothing is in flight to carry it.
 #
 # block-stop-on-unsettled-pr.sh returns 0 before looking at anything when the open-pull-request
-# list is empty. That is the state this fires in, and it is the state the failure happened in:
-# six issues open, a next action named in the closing paragraph, and the turn ended without it.
-# Naming the next action is what the stall looks like from inside — it reads as a plan rather
-# than as stopping — so the guard cannot key on intent. It keys on the backlog being non-empty
-# with nothing carrying it.
+# list is empty. That is the state this fires in, and it is the state the failure happened in: a
+# backlog of assigned issues, the next action named in a closing paragraph, and the turn ending
+# without it. Naming the next action is what the stall looks like from inside — it reads as a plan
+# rather than as stopping — so the guard cannot key on intent. It keys on the backlog being
+# non-empty with nothing carrying it.
 #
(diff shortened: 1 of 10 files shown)


s4k10503:
Replaced by the branch this was rebased onto after #323 merged. Same three commits plus the anchor-guard fix.

PR #326, opened 2026-08-01
Hold the region strip's width against the tree's own preprocessor conditions

### LINKED ISSUE #310: The doc guard's widening direction is held by one sentence in an unguarded file
`DocumentationDriftTests` strips `#region`/`#endregion` lines from the identifier corpus. Two directions can break:

- **the strip stops running** — held by `Given_TheRepoSources_When_TheIdentifierCorpusIsBuilt_Then_TheRegionStripTakesOnlyLabelWords`, which rebuilds the corpus without the branch and fails on an empty difference;
- **the strip widens** to every preprocessor directive — held by nothing in the fixture, because the control and the label scan read the same shared constant.

The comment on that case says the widening is caught by the identifier check instead, and today it is. Measured: widening the branch to `^[^\S\n]*#[^\n]*$` drops exactly eight words from the corpus — `CS0649`, `ENABLE_IL2CPP`, `NET5_0_OR_GREATER`, `UNITY_EDITOR`, `annotations`, `endif`, `pragma`, `warning` — and only `UNITY_EDITOR` is cited anywhere in the scanned markdown.

Both citations are in one sentence, `Packages/com.velvet.core/Generators~/README.md`, the bullet about compiling with and without the editor define.

So a rewrite of that bullet — an edit with no visible connection to the doc guard — retires the widening guard silently. The precedent is on record: #297 removed the backticks from `` `Group D` `` in `styling-filters.md` to make a guard pass, so a load-bearing backtick span disappearing to satisfy a check is a move this repo already makes.

## What would settle it

A case that asserts the corpus keeps a known set of directive-condition tokens, derived from the sources rather than from a guide's prose. `UNITY_EDITOR`, `ENABLE_IL2CPP` and `NET5_0_OR_GREATER` are all real conditions in the tree, so the set is available without citing anything.

Not urgent — the guard is correct today. Filed so the dependency is written where it can be acted on rather than only in the comment that depends on it.

s4k10503:
Closes #310.

`DocumentationDriftTests` strips `#region` and `#endregion` labels from the identifier corpus. Only one direction of that strip is held: `Given_TheRepoSources_When_TheIdentifierCorpusIsBuilt_Then_TheRegionStripTakesOnlyLabelWords` fails when it stops running.

The other direction — widening it to every preprocessor directive — was caught only as a side effect. Widening drops the tree`s condition tokens out of the corpus, and `Generators~/README.md` happens to cite `UNITY_EDITOR` in a backtick span, so the identifier check reports it. A rewrite of that one bullet would retire the guard with nothing saying so, and this repository has already removed a load-bearing backtick span to make a check pass.

The new case reads `#if` and `#elif` conditions out of the raw sources and asserts the corpus keeps every one. Measured in the tree today: `UNITY_EDITOR` in 60 conditions, `ENABLE_IL2CPP` in 5, `NET5_0_OR_GREATER` in 1. It derives them rather than listing them, so a condition added or dropped needs no edit here.

The conditions are read from raw text while the corpus is built from stripped text, which is what makes the two sides disagree under a widened strip rather than widen together. The count of conditions found is folded into the assertion, so a walk that read no directive at all cannot satisfy it.

## Verification

- Green: the fixture`s 8 cases pass unmodified.
- Red: widening `RegionLabelAlternation` to `^[^\S\n]*#[^\n]*$` fails the new case. The incidental identifier check fails alongside it, which is the dependency this replaces rather than removes.

The comment on the region case no longer claims the widening is caught by the identifier check, since it now names the case that holds it.

No `Documentation~` impact: this adds a guard over the guides rather than changing anything they describe.

### DIFF
--- a/Packages/com.velvet.core/Runtime/Component/Tests/Editor/DocumentationDriftTests.cs
+++ b/Packages/com.velvet.core/Runtime/Component/Tests/Editor/DocumentationDriftTests.cs
@@ -235,11 +235,44 @@ namespace Velvet.Tests
 
             // Assert — a strip that is not running removes nothing, so an empty first term is the shape
             // that catches it. The second catches a strip widened by any route other than the spelling both
-            // sides read; widening that spelling to every directive is caught instead by the identifier
-            // case, and only for as long as a guide keeps citing UNITY_EDITOR — Generators~/README.md is
-            // the one that does.
+            // sides read; widening that spelling to every directive is caught by the case below.
             Assert.That(
                 (removed.Count > 0, string.Join(", ", notFromALabel)),
+                Is.EqualTo((true, string.Empty)));
+        }
+
+        [Test]
+        public void Given_TheRepoSources_When_TheIdentifierCorpusIsBuilt_Then_ItKeepsEveryDirectiveCondition()
+        {
+            // Arrange — the conditions are read from the raw text while the corpus is built from the
+            // stripped text, so a strip that widens from the two region labels to every directive takes
+            // them out of one side and not the other. Deriving them rather than listing them is what keeps
+            // this from being a second copy of the strip: nothing here has to be updated when a condition
+            // is added or dropped.
+            var directive = new Regex(@"^[^\S\n]*#[^\S\n]*(?:if|elif)\b([^\n]*)$", RegexOptions.Multiline);
+            var word = new Regex(@"[A-Za-z_][A-Za-z0-9_]*", RegexOptions.Compiled);
+            var conditions = new HashSet<string>(StringComparer.Ordinal);
+            foreach (var entry in DocumentationCorpus.RepoEntries(includeClaude: false).Where(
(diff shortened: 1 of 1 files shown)


s4k10503:
Rebasing onto the main that now carries #332 broke this: that change moved the repository walk from a private field in this fixture to `DocumentationCorpus.RepoEntries(bool)` and rewrote every call site it could see. The call site added here was not one it could see, so the rebase applied cleanly on separate lines and left `RepoEntries.Value` naming nothing.

Two branches, each green, neither containing the other change — the shape `refuse-stale-merge.py` exists for. It refused the merge, the rebase surfaced the break, and CI reported it. Fixed by pointing the new case at the shared walk with `includeClaude: false`, the same argument every other case in this fixture passes.

Whole EditMode suite after the fix: 3585 cases, 3582 passed, 0 failed, 0 inconclusive, no compile errors.

PR #359, opened 2026-08-02
fix(velvet): audit what the commit records rather than the working tree

### LINKED ISSUE #354: The commit guard audits the working tree rather than what the commit records
`refuse-commit-failing-fast-checks.py` refuses `git commit` when a staged file fails a fast deterministic check. Six ways it does not.

## It reads the working tree, not what the commit records

`check_python`, `check_shell`, `check_json` and `check_yaml` open `os.path.join(root, rel)`. The commit records the index blob. Stage a broken file, fix the working copy, and the commit passes every check while `git show :m.py` still holds `def broken(:`.

The same split has two more faces:

- A file staged for addition and then deleted from the working tree is refused with a raw `FileNotFoundError` traceback, for a commit git would accept.
- `git commit -a` and `git commit <pathspec>` are recognised as commits but audited against the index only, so the working-tree content those forms actually commit is never inspected — a live sweep neuter among it, which is the state `staged_neuters` exists to catch.

`staged_neuters()` in the same file already reads `git show :<file>`, so the two halves of this hook disagree about what a commit is.

## A staged rename is checked by nothing

`staged_paths` filters `--diff-filter=ACM`. A rename reports `R`:

```
git mv good.py moved.py     # then break moved.py and stage it
git diff --cached --name-status   ->  R059  good.py  moved.py
git diff --cached --name-only --diff-filter=ACM   ->  (nothing)
```

The hook returns 0. It needs git to report `R`, so the deletion of the old path must be staged too and similarity must clear the rename threshold.

## An unparseable cut file skips every check

`cut_targets()` calls `json.load` unguarded. With `scripts/neuter-cuts.json` malformed the hook exits 1 with a traceback, PreToolUse treats 1 as non-blocking, and the fast checks and the neuter check are both skipped — the failure mode is the guard turning itself off.

## `-C` is read for recognition and ignored for the audit

`git -C <dirty> commit` run from a clean tree passes; `git -C <clean> commit` run from a dirty tree is refused, naming a file in the other repository. Both halves need the path unquoted: a quoted `git -C "$DIR" commit` is not recognised as a commit at all, because the mask blanks the argument before `COMMIT.search` runs.

## Shape of the fix

The audit should read the content the commit will record — index blobs for a plain commit, plus the working tree for the forms that commit it — and run against the repository the command names. The command recognition wants the same tokenisation the branch guard now uses, so the parsing belongs in `.claude/hooks/lib/`.

s4k10503:
Closes #354.

The hook opened `os.path.join(root, rel)` — the file in the checkout — while a commit records the index blob. Those are different bytes, and the gap runs both ways.

## Measured, against the version on main

Each case in a throwaway repository, driving both hooks with the same event:

| case | main | this |
| --- | --- | --- |
| broken blob staged, working copy fixed afterwards | exit 0 | **exit 2** |
| staged rename carrying a syntax error | exit 0 | **exit 2** |
| `git commit -am` with a broken working-tree file | exit 0 | **exit 2** |
| malformed `scripts/neuter-cuts.json` | exit 1, traceback | **exit 2** |
| staged for addition then deleted from the tree | exit 2, `FileNotFoundError` | **exit 0** |

The last row is the same defect seen from the other side: a commit git accepts, refused with a traceback instead of a syntax report.

The fourth is the worst-shaped of them. `cut_targets()` called `json.load` unguarded, so a malformed cut file raised out of `main` — and a PreToolUse hook exiting 1 is non-blocking, so the one file whose breakage matters most turned every check off and let the commit through.

## What it audits now

The content the commit would record: index blobs for a plain commit, and the working tree as well for the forms that commit it. `--diff-filter` keeps `R`, so a rename is no longer invisible. `git -C <dir> commit` audits `<dir>` rather than the event's cwd, which named one tree and measured another.

`py_compile` and `shellcheck` want a file, so the content is written to a scratch file and the report names the repository-relative path.

The neuter check moved onto the same content. `git commit -am` was the form that could carry a live sweep's neuter past it, which is what that check exists to stop.

## Shared parsing

Recognising the command wants the tokenisation the branch guard now uses — quoting defeated `-C` here too, so `git -C "$DIR" commit` was not recognised as a commit at all. That parsing moves to `.claude/hooks/lib/shell_commands.py` and both hooks import it. The branch guard's own table is unchanged by the move: `BranchGuardParsingTests` still passes on all 33 rows.

`CommitGuardParsingTests` pins 18 commands and, for each, the three answers the audit is driven by — repository, whether the working tree is committed, and pathspecs — since a misreading there audits the wrong bytes rather than merely missing a refusal. Short-flag bundling is in the table because the first attempt read `git commit -am wip` as committing a pathspec named `wip`.

## Verification

Whole EditMode suite on this branch: **3588 passed, 0 failed, 0 inconclusive**, no compile errors.

## Documentation

No `Documentation~` impact: the guard is harness tooling and `CONTRIBUTING.md` does not describe its command surface.

### DIFF
--- a/.claude/hooks/lib/shell_commands.py
+++ b/.claude/hooks/lib/shell_commands.py
@@ -1,0 +1,184 @@
+"""Shell-command parsing shared by the hooks that must recognise a git invocation.
+
+A regex over the masked command missed eleven spellings of a branch creation at once, quoting the
+argument among them, and every miss was silent. Splitting into segments and tokenising is what
+those hooks agree on; each keeps its own table of which subcommand and which flag it cares about.
+
+The mask locates command boundaries, where it is correct, and the tokens come from the original
+text, because a masked argument has been replaced by spaces and cannot be read.
+"""
+
+import os
+import re
+import shlex
+
+SEPARATORS = set(";&|\n")
+
+# Words that may precede the command without changing which command it is. `then`/`do`/`else`
+# because a command inside a conditional or a loop is still that command.
+LEADING_WORDS = {"then", "do", "else", "elif", "!", "time", "command", "nohup", "exec"}
+ENV_ASSIGNMENT = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*=")
+REDIRECTION = re.compile(r"^\d*(?:>>|>&|<&|>|<)")
+
+# git's own options, before the subcommand. Only -C is returned: it names the repository the
+# command acts on, and evaluating the cwd instead answers about a different tree.
+GLOBAL_VALUE_FLAGS = {"-C", "-c", "--git-dir", "--work-tree", "--namespace", "--config-env"}
+
+
+def mask_shell_literals(command):
+    out = list(command)
+    i = 0
+    n = len(command)
+    while i < n:
+        ch = command[i]
+        if ch in "'\"":
+            quote = ch
+            out[i] = " "
+            i += 1
+            while i < n:
+                if quote == '"' and command[i] == "\\" and i + 1 < n:
+                    out[i] = out[i + 1] = " "
+                    i += 2
+                    continue
+                if command[i] == quote:
+                    out[i] = " "
+                    i += 1
+                    break
+                out[i] = " "
+                i += 1
+        elif ch == "\\" and i + 1 < n:
+            out[i] = out[i + 1] = " "
+            i += 2
+        elif command.startswith("<<", i):
+            j = i + 2
+            strip_tabs = j < n and command[j] == "-"
+            if strip_tabs:
+                j += 1
+            if j < n and command[j] in "'\"":
+                quote = command[j]
+                j += 1
+                start = j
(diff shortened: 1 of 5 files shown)


s4k10503:
One more, found by the guard refusing a commit it should not have.

`shellcheck` exits non-zero on info-level notes as well as on warnings and errors, and three hooks in this repository carry one — SC1091, for sourcing a sibling that was not passed as an input. So every commit touching `block-shared-git-state.sh`, `block-stop-on-open-backlog.sh` or `block-stop-on-unsettled-pr.sh` was refused for a note nobody intends to act on.

The check is now floored at `--severity=warning`. Measured either side: a staged file whose only finding is SC1091 gives exit 0, a staged file with a shell syntax error still gives exit 2.

Whole EditMode suite with the floor in place: 3588 passed, 0 failed, 0 inconclusive.

PR #377, opened 2026-08-05
refactor(velvet): replace UniTask with an in-tree VelvetTask

### LINKED ISSUE #375: Drop the UniTask dependency
## Goal

Velvet ships with no dependency on UniTask: `Runtime/Velvet.asmdef` stops referencing the `UniTask` assembly, no public signature names a `UniTask` type, and installing Velvet no longer requires the consumer to install anything alongside it.

## Why

UniTask is the one peer dependency a consumer has to install by hand (`README.md` — "a required peer dependency you install yourself"), which makes installation a two-step operation and forces every consuming project to carry Cysharp's assemblies whether or not it writes async code of its own. A framework asking for a virtual DOM should not also dictate the project's async type.

## Where the dependency sits today

Surveyed on `main` (2026-08-05):

- Assembly reference: `Packages/com.velvet.core/Runtime/Velvet.asmdef` references `UniTask` by assembly name.
- Runtime files naming a UniTask type — 11, none of them optional:
  - `Hooks/Hooks.cs`, `Hooks/HookMutation.cs`, `Hooks/HookSlots.cs`, `Hooks/TransitionStarter.cs`
  - `Routing/Router.cs`, `Routing/RouteDefinition.cs`, `Routing/RouteBlockerManager.cs`, `Routing/RouteLoaderRunner.cs`
  - `Reconciler/FiberAsyncResource.cs`, `Reconciler/FiberWorkLoop.cs`
  - `Component/V.cs`
- Public surface: 17 public declarations name a UniTask type — among them `Hooks.Use<T>(Func<UniTask<T>>)`, `Hooks.UseNavigate`, `Hooks.UseBlocker`, `Router.NavigateAsync`/`GoBack`/`GoForward`, `RouteDefinition.Loader`, `V.Route(loader:)`, `MutationResult.MutateAsync`, `TransitionStarter.Invoke(Func<UniTask>)`. This is a breaking API change for consumers, not an internal swap.
- Test and tooling surface: the Editor test asmdefs reference `UniTask`, and the generator suite carries a `UniTask<T>` stub (`Generators~/tests/.../GeneratorTestHelper.cs`) that `StubSurfaceDriftTests` pins to the real signatures — a change here has to move with the generator suite.

## Constraints any replacement has to hold

These are the properties the current code depends on; a replacement that drops one is a regression, not a migration.

1. **Zero-allocation await on the steady path.** Async is on the render/navigation path, so per-await state-machine allocation is a per-frame cost. Measure with the GC.Alloc recorder (`TestUtilities/GCAllocationProbe`), not `GC.GetAllocatedBytesForCurrentThread`.
2. **Continuations resume on the main thread, driven by the PlayerLoop** — not by whatever `SynchronizationContext` happened to be captured.
3. **Double-await safety.** `FiberAsyncResource` peeks a task's status, stores the task, and awaits it later; `RouteLoaderRunner` collects tasks into a `List` and awaits them after the loop. A pooled awaitable that reuses instances without a version token turns both of those into silent corruption rather than an exception. Whatever the replacement is, these two call sites are the ones to design against.
4. Equivalents for the helpers currently in use: `UniTask.FromResult`, `.Forget()` with fire-and-forget exception routing, `.AttachExternalCancellation(token)`, and a synchronously-completed status peek.

## Options to evaluate

Both of the obvious native candidates were prototyped and rejected once before; the rejection was on constraints 1–3 above rather than on features, so re-opening it means producing measurements, not opinions.

1. **`UnityEngine.Awaitable`** — PlayerLoop-native, but pooled instances are not safe to await twice ([ruccho](https://zenn.dev/ruccho/articles/c0b657c9e17b7e), [UniTask discussion #627](https://github.com/Cysharp/UniTask/discussions/627)), and constraint 4's helpers do not exist. Supplying both is close to reimplementing UniTask in-tree.
2. **`ValueTask`** — safe to await twice via the BCL version token, but resumption depends on the captured `UnitySynchronizationContext` and the async state machine still allocates.
3. **A Velvet-owned awaitable** — an `IValueTaskSource`-style pooled source driven from Velvet's own PlayerLoop hook. Satisfies all four constraints by construction, at the cost of owning a concurrency primitive in this repo forever.
4. **Keep UniTask internally, remove it from the public surface** — does not reach the goal (the assembly reference and the install step stay), but scopes what a consumer has to know. Worth pricing as the fallback.

## Done when

- `Runtime/Velvet.asmdef` has no `UniTask` reference, and no public signature names a UniTask type.
- `README.md` and `Packages/com.velvet.core/README.md` no longer instruct the consumer to install UniTask; installation is a single package add.
- EditMode + PlayMode suites are green, with `Routing`, `Reconciler` (Suspense) and `Hooks` (`Use`, `UseMutation`, `UseTransition`) fixtures exercising the new type.
- A benchmark shows no allocation regression on an await-per-frame path, run against a fixture that actually drives the changed code.
- The affected `Documentation~` guides are updated in the same PR.

s4k10503:
Velvet no longer depends on UniTask. `Runtime/Velvet.asmdef` and every test asmdef drop the assembly
reference, the package manifests no longer declare `com.cysharp.unitask`, and installing Velvet is a
single package add — the READMEs say so.

## What replaces it

`VelvetTask` / `VelvetTask<T>` under `Runtime/Async/`, with the surface the codebase actually uses:
`Forget`, `FromResult`, `AttachExternalCancellation`, completion sources, `Yield`, `Never`,
`FromException`, `CompletedTask`, `ToCoroutine`, and a synchronous status peek. No `WhenAll`, no
`Delay`, no async LINQ — those had no call sites.

The design follows UniTask's: a pooled source carrying a version token, a custom async method builder
that stays lazy until first suspension, and completed tasks held inline in the struct so a
synchronously-completed path allocates nothing. Frame-bound continuations are driven from the
PlayerLoop at runtime and from `EditorApplication.update` in the editor, so `async VelvetTask` methods
resume in EditMode — which is where the EditMode suites exercise them.

`UnityEngine.Awaitable` and a BCL `Task` need no adapter: both builders declare the two awaiter
constraints the C# compiler emits an await against, so either is awaited as it stands inside an
`async VelvetTask`. `VelvetTaskAwaiterInteropPlayModeTests` pins that, and the resumption thread the
`Task` path takes. `Documentation~/async.md` states it.

## Breaking change

The CHANGELOG entry sits under `## [Unreleased — breaking]`, beside the fifteen breaking entries
already waiting for the next major.

Every public async signature changes type. Consumers update their own loaders, blockers and mutation
functions:

```csharp
// before
Func<RouteLoaderContext, CancellationToken, UniTask<object>> loader;
Func<NavigationAttempt, CancellationToken, UniTask<bool>> shouldBlock;
async UniTask<Data> Mutate(Vars v, CancellationToken ct);

// after
Func<RouteLoaderContext, CancellationToken, VelvetTask<object>> loader;
Func<NavigationAttempt, CancellationToken, VelvetTask<bool>> shouldBlock;
async VelvetTask<Data> Mutate(Vars v, CancellationToken ct);
```

Consuming a task twice throws `InvalidOperationException` rather than returning reset state. Tasks
completed inline (`FromResult`, and an `async` method that never suspends) may be read more than once,
matching UniTask.

## Allocation

Measured in PlayMode, warmup 5 / 20 measurements, with both instrument canaries live, on the tree as it
stood before this branch took `main`'s rewrite of `RouteLoaderRunner.cs`:
GC.Alloc blocks are 12 for `Hooks.Use` suspend/resume, 14 and 11 for the two `RouteLoaderRunner` paths,
98 for `Router.NavigateAsync` with a loader and a blocker, and 22 for a `UseMutation` round trip.

**There is no before column, and this change should not be read as a performance change.** The UniTask
numbers this branch was opened with were measured against a `main` 116 commits old, and four of the five
paths have been rewritten there since — the suspend loader's round accounting, `UseMutation`'s generation
ownership, and the router's blocker resume and publishing root. A before-and-after over two different
bodies of code answers nothing, and re-taking the UniTask side would mean porting these fixtures onto a
base that no longer exists.

What this change is for is the dependency, not the speed.

## What the struct path costs to reach at all

Editor compilation emits **class** state machines, so neither suite reaches the value-type path on its
own and the one hand-rolled fixture covered a single builder overload. Three more cover the rest — the
non-generic `AwaitOnCompleted` and both `AwaitUnsafeOnCompleted`, which is the pair Roslyn emits for
`await someTask`, and so the shape the interop section of the guide describes. Each is red against the
pre-fix `Rent` and green after, one cut to one case over a 4170-case suite.

`Documentation~/async.md` gains two things a reader needs and did not have. A frame-spanning test is
driven from a `[UnityTest]` through `VelvetTask.ToCoroutine`, which is public, rather than through a
drain helper that is internal to the test utilities — a synchronous `[Test]` body never returns to
either hook between the yield and the assertion. And a continuation resumed off the main thread must
not call back into Velvet: the pools it rents task sources and runners from, and the frame driver's
queues, are plain collections that nothing synchronizes, and `VelvetTask.Yield()` is not a way back —
scheduling one appends to the list the main-thread drain swaps and clears.

## Docs

`Documentation~/async.md` covers the type and its EditMode behaviour, with the index updated.
`Documentation~/react-migration.md` and both READMEs no longer reference UniTask. The CHANGELOG
records the breaking change under `[Unreleased — breaking]`.

Closes #375


s4k10503:
Correcting what this was parked behind, because the stated reason has stopped being true.

This was being held as "the 3.0.0 gate", waiting on 2.1.1 to publish. **It is not the gate.** `## [Unreleased — breaking]` on `main` now carries eleven entries across `### Added`, `### Changed` and `### Fixed`, so the next release cut from `main` is a major whether or not this lands. This branch can go into 3.0.0 or into any later major; nothing about the version number waits on it.

What it actually waits on:

- **A 112-commit rebase over 146 files.** Last updated 2026-08-06; `origin/main` has moved 112 commits since. `mergeStateStatus` is DIRTY, so it cannot go green on its own, and the conflict surface is the whole diff rather than a few files.
- **A decision that the replacement is still wanted.** Retaining UniTask and rejecting `Awaitable`/`ValueTask` was settled separately; an in-tree `VelvetTask` is a third answer and has not been re-argued against the tree as it stands 15 days on.

Both are real costs and neither is urgent while the backlog holds product bugs. Leaving it a draft is the right state; what was wrong was the reason attached to it.

s4k10503:
Held as a draft on its mutation campaign, which is queued behind #800's — the campaign needs an editor per mutant, so only one runs at a time here.

The campaign was previously measured as unable to pass regardless: 16 of 122 mutants timed out, all of them removing what completes an await, and a timeout was neither a kill nor an answer. #818 has since landed, which reads a mutant that hung as killed where the baseline's own wall clock leaves room to tell the two apart. That is what makes the run worth spending.

s4k10503:
EditMode is green locally now — 4232 passed, 0 failed. Three things had drifted:

- `BlockerTests` named `UniTask.FromResult` in a case main added after the last merge. Nothing
  conflicted, because main's line compiles on main; it is this branch that has no such type.
- `PublicAPI.txt` still described `VelvetTask<T>` as its own awaiter. The code moved to a nested
  `Awaiter<T>` struct and the file was never regenerated.
- The two navigation allocation pins moved from 98 to 99. The block is the splat-position check
  main gained in #862: it walks the route's own segments through an iterator, once per route that
  declares children, and `CreateRouter` is inside the measured call.

What is left is not the branch's. `base-red` reports 280 cases as `the base tree cannot answer`,
because the base has no source for `VelvetTask` and every carried test module names it. The local
withdraw loop reaches the same verdict — EditMode repeated the same round eight times without
withdrawing anything further, four seconds each, which is the compile abort rather than a run.
Filed as #865 with the measurement.

s4k10503:
Rebased onto `main` now that #878 has landed the type. The tree compiles and 4246 EditMode cases pass;
two fail, and they are the same two, with a finding under them.

`RouterNavigationAllocationEditorTests`' pins read 104 where they pin 99. Bisected by deleting one
line of the fixture: with `router.RouteBlockerManager.Register(...)` removed, both pass. **All five
extra blocks come from the blocker path**, not from navigation or the loader — the loader is
`LoaderMode.Await` returning `VelvetTask.FromResult`, and it costs nothing new.

That is worth understanding before the number moves. This branch's own claim is that a
synchronously-completed path allocates nothing, and a blocker returning `VelvetTask.FromResult(false)`
is exactly such a path: `RouteBlockerManager.CheckAsync` awaits it and never suspends, so the builder
should stay lazy and the source should never be rented. Five blocks say something on that path is
allocating that its UniTask equivalent did not.

I have not moved the pins. The comment above them says moving the number is the answer when the
change that moved it is understood, and this one is not yet.

The rest of the rebase: `SuspenseBoundaryTests` gained a case on `main` that names
`UniTaskCompletionSource`, and `UseAsyncTests` / `RouterTests` took `main`'s side of three conflicts
with the type swapped — the improvements there are `main`'s and the type is this branch's.

s4k10503:
Green locally: EditMode 4250 passed / 0 failed, PlayMode 193 / 0.

The allocation pins moved from 99 to 101, and every block between the 104 they read on the last rebase
and the 101 they read now is accounted for. Three of the five were regressions on `main`, each found
by these pins and fixed there first:

- #881 — `Match` allocated the record of which segment each pattern segment took once per branch it
  walked past.
- #882 — resolving a match's base re-parsed each route's path through an iterator, once per level.
- #883 — a blocker pass allocated a snapshot array on every navigation.

The remaining two are this fixture's own shape. It builds a router inside the measured call, so it
pays what a long-lived router amortises: #883's reusable snapshot is two blocks the first time and
none after, where the array it replaced was one every time. A navigation on a router that already
exists is cheaper than before; the first one on a fresh router is one block dearer, and this fixture
measures only first ones.

That is why the pins move rather than the code: the comment above them asks for the change that moved
the number to be understood, and now it is.

s4k10503:
Every check passes except the two base-red lanes, and after the split the reason is precise.

The additive half is now on `main` (#878), so the base has `VelvetTask`. What this half changes is the
type of Velvet's own signatures — `Router.NavigateAsync` returning `VelvetTask<NavigationResult>`
where the base returns `UniTask<NavigationResult>`. A carried test module that names the method
compiles against neither tree's other spelling, and the name it spells is one **both** trees have. That
is the case `base_red_check`'s module docstring names as the one its static comparison cannot reach:

> What the comparison cannot reach — a signature the branch changed under a name both trees spell —
> leaves a run that measured nothing, which fails.

212 cases report `the base tree cannot answer`. Splitting further does not help, and neither does the
round budget: the withdraw loop reaches an answer by putting every affected module back, and every
module is affected. Nor can the migration be staged behind an overload, because C# does not overload
on return type.

So this is not a defect in the branch and not one a further split can lift. What remains is the
judgement I have no standing to make: whether a change base-red is structurally unable to read merges
with those two checks red. #865 records the same conclusion from the tooling side, including the two
fixes I made and the one I measured and rejected.

Local state: EditMode 4250 passed / 0 failed, PlayMode 193 / 0, on `main` as of #883.

s4k10503:
## Correction: step 1 of the plan above is not available

"Production keeps both types" cannot be done for the members that matter. Measured with the
compiler, not argued:

```csharp
[AsyncMethodBuilder(typeof(AsyncTaskMethodBuilder))] public struct A { public TaskAwaiter GetAwaiter() => default; }
[AsyncMethodBuilder(typeof(AsyncTaskMethodBuilder))] public struct B { public TaskAwaiter GetAwaiter() => default; }

public static void Use(Func<A> f) { }
public static void Use(Func<B> f) { }

Api.Use(async () => { await Task.Yield(); });   // CS0121: the call is ambiguous
```

An `async` lambda infers its delegate type from the target, and with two awaitable-returning
candidates there is no target to infer from. Adding a `VelvetTask` overload beside each `UniTask`
one therefore breaks every existing call site rather than leaving it alone — and the call sites are
the point, because 24 of the 41 `UniTask` positions in `Runtime/` are parameters of the shape
`Func<…, UniTask<T>>`: `Hooks.Use`, `Hooks.UseBlocker`, `UseMutation`'s `MutationFn`,
`V.Route(loader:)`. Those are application callbacks, and an application writes them as `async`
lambdas.

## What that costs the split

Without step 1, a test module migrated to the new signature names a member the base does not have in
that shape, so it does not compile there. The static withdrawal cannot see it — `VelvetTask` is a
name both trees spell, and `added_csharp_surface` compares spellings — so CI's single round compiles
nothing however few modules the pull request carries. Splitting reduces how many rounds the local
loop needs; it does not make one round enough.

So the split does not reach a base-red CI can answer, and the two options left are the ones already
on the table: a receipt CI checks against the tree, or merging with those two checks red. Neither is
mine to choose, and the plan above should not be started as written.

s4k10503:
## Why the two base-red lanes are red, and what was measured

`Base-red check (EditMode)` refuses, and names its own remedy:

```
212 case(s) yielded no base verdict, so none of them carries a reading either way. A
declaration does not answer for that -- the detail beside each says what happened.

The editor wrote nothing, which a single round cannot tell from an editor that never
started. Take the reading where the loop runs -- it withdraws by the editor's own error
list and asks again until something answers:
  python3 scripts/test_quality/base_red_check.py --lane csharp --platform EditMode --platform PlayMode --base 7c446b78 \
    --warm-library Library
```

Taken, at `--max-rounds 40`:

```
EditMode attempt 27: 0 case(s) over 3 fixture(s) in 3s
no round wrote a result, so nothing any of them was asked was measured
```

**The loop ran out of files it could withdraw before it ran out of budget** — 27 rounds of 40, three
fixtures still carried, and the base compiled nothing at any point. What the last round was left
holding was one code:

```
123 CS1929
```

`CS1929` is a receiver type an extension method does not take: the base's `UniTask` does not accept
the extensions this branch's tests call on `VelvetTask`. Round 1 by contrast raised `CS0234`,
`CS0246` and `CS0616` — names the base has not got — and those are what the withdrawals cleared.

That is the difference that decides this. A withdrawal answers a name the base lacks: the file comes
out and the rest of its assembly builds. It answers nothing here, because the file that fails is the
file each case is in — putting it back removes the case rather than the obstacle.

## So there is nothing to sharpen

The question base-red asks is whether a case would have failed without this change. These cases
cannot be written against the base at all, and not because they are new: the base holds the same
members under the shapes this replaces. More rounds do not reach it, which was measured rather than
assumed — the remedy that claimed otherwise was withdrawn in #901.

Everything else is green: EditMode 4259 passed / 0 failed, PlayMode 193 / 0, and the other eleven
checks pass.

Merging this means merging with those two lanes red, for the reason above. `settle.py` will not do
it and branch protection will not either, so it needs a decision and an action that are not mine to
take.

s4k10503:
## Rebased, and the red is now only the two lanes that cannot answer

Five commits on `origin/main`, no merge commit, tree byte-identical to what was measured before the
rebase. Verified on the rebased tree with ~40 commits of `main` in it: **EditMode 4260 passed / 0
failed, PlayMode 193 / 0**, no compiler error.

The Python lane's failure is gone with it. That one was the branch trailing `main` — it did not carry
the release scripts `main` has gained — rather than anything about this change.

What is left is what the comment above measures: `Base-red check (EditMode)` and `(PlayMode)`, and
`Required checks (Unity)`, which follows from them.

Nothing here changes the reading. The loop still runs out of files it can withdraw before it runs out
of budget, and what it is left holding is one code — `CS1929`, a receiver type an extension method
does not take. There is no case to sharpen, and a bigger budget does not reach it.

s4k10503:
Bringing `main` in (it was seven commits behind) changed what the red is, so the reason posted earlier
is stale. What the two C# lanes now report is not a case that passed on the base — it is that the base
tree did not run at all:

    the base tree cannot answer   Velvet.Tests.DocumentationDriftTests.…  (the run wrote no results file, so nothing was measured)

repeated for every planned case. The cause is in the job log and not in the verdict:

    Aborting batchmode due to failure:
    Scripts have compiler errors.
    …/FiberAsyncResourcePlayModeBenchmarks.cs(50,30): error CS0012: The type 'UniTask<>' is defined in
      an assembly that is not referenced.
    …/UseMutationPlayModeBenchmarks.cs(41,17): error CS0012: …

The static withdrawal ran first and took four `Routing/Tests/PlayMode` files out for an asmdef the base
has not got, so it is doing its half. What it cannot reach is this: `base_red_check` carries `.cs` files
only — its own comment says "the asmdef is not a `.cs` and so is never carried" — and this branch
removes a reference from three asmdefs. A base tree assembled from the base's asmdefs and whichever
`.cs` files were carried can then be a pairing neither side has.

I have not established which pairing produces these two errors. Both files are changed by this branch
and neither branch version mentions `UniTask` at all, so the base tree is holding the base's copy of
them — which means the answer is in what the base tree looked like, and CI discards it.

**This is not something I can settle from here.** The branch is otherwise green — EditMode, PlayMode,
player compilation, generators, test quality and both release jobs all pass on the current head — and
`settle.py` refuses a red merge while branch protection reports `BLOCKED`, so merging it is an admin
call rather than a fix.

Recorded against #635, which is exactly this shape: the verdict lane gets `plan.json` and the results
directory and no log, so it prints "the run wrote no result" and stops. Here that costs a reader the
two lines that say what actually happened.

s4k10503:
Settled it by building the base tree here and running it, which is the reading CI discards.

    453 compile errors, and the first of them:
    …/UseMutationPlayModeBenchmarks.cs(41,17): error CS0012: The type 'UniTask<>' is defined in an
      assembly that is not referenced. You must add a reference to assembly 'UniTask, …'

The mechanism, each step measured rather than inferred:

1. This branch **adds** those PlayMode benchmark files — `git show <base>:…UseMutationPlayModeBenchmarks.cs`
   is "exists on disk, but not in <base>".
2. `base_red_check` carries them onto the base tree, because they are test-side `.cs`.
3. Line 41 is `s_captured.MutateAsync(21).GetAwaiter().GetResult()`. On the base, `MutateAsync`
   returns `UniTask<T>`, so resolving `GetAwaiter` needs the UniTask assembly.
4. The asmdef that owns them is the **base's** — asmdefs are never carried, and `base_red_check` says
   so itself — and it references `Velvet`, `Velvet.TestUtilities`, the test runners and
   `Unity.PerformanceTesting`. No UniTask.
5. So the tree cannot build, no results file is written, and every planned case reads "the base tree
   cannot answer".

`unbuildable_on_base` cannot see this. It withdraws a carried file that spells a **name** the base has
not got — which is why the four `Routing/Tests/PlayMode` files were withdrawn correctly. Here every
name exists on the base; what is missing is an assembly *reference* for a type that appears only in the
base's own signature.

**So this is not fixable on the branch in any way worth doing.** The documented workaround is to reach
new API by reflection, and these are allocation benchmarks whose whole subject is the call — rewriting
them through reflection would measure the reflection. The alternative is to delete them from the
branch, which removes coverage this change is supposed to add.

The branch is otherwise green: EditMode, PlayMode, player compilation, generators, test quality and
both release jobs all pass on the current head. `settle.py` refuses a red merge and branch protection
reports `BLOCKED`, so landing it is an admin call.

Recorded on #635, which is the harness side of this: the verdict step gets no log, so none of the
above reaches the reader who needs it.

Issue #462, opened 2026-08-06
An explicit null deps list freezes UseCallback and UseMemo but re-runs UseBlocker and UseEffect


s4k10503:
Passing `null` as a hook's dependency list means opposite things depending on which hook you pass it to, with nothing in the signatures or the docs to distinguish them.

**Freeze** — the value is computed once and never recomputed:

- `Hooks.UseCallback(cb, null)`
- `Hooks.UseMemo(f, null)`

Both compare `ObjectIs.AreEqualDeps(entry.LastDeps, deps)` with no null guard, and `AreEqualDeps(null, null)` returns `true` at its `ReferenceEquals` line.

**Re-run** — every render:

- `Hooks.UseBlocker(pred, null)`
- `Hooks.UseEffect(f, null)`

Both guard on `deps != null` before the comparison.

A user who learns the behaviour from one and applies it to the other gets the opposite outcome, silently. The freeze direction is the dangerous one: a `UseCallback` that never updates captures stale state, which is the classic React bug the exhaustive-deps analyzer exists to prevent, and here it is reachable by writing `null` rather than by omitting a dependency.

Note this is separate from *omitting* the list, which is well-defined for every one of these — the deps-less overloads mean "every render" consistently. The divergence is specifically about an explicit `null`.

Whichever way it is unified, the deps-taking overloads should agree, and the XML docs should state what the null case does — today only `UseAnimationSequence` and (after #398) `UseBlocker` do.

Issue #461, opened 2026-08-06
Hooks.UseContext promises non-null while returning a null default, so reading a router context off it throws with no warning


s4k10503:
`Hooks.UseContext<T>` is declared to return `T` — non-nullable — and its body ends in `context.DefaultValue!`, while `ComponentContext<T>.DefaultValue` is declared `T?`. `RouterContext.Location` and `RouterContext.OutletContext` are both created with `Create(null)`.

That `!` is what suppresses CS8603 on the way out. So:

    Hooks.UseContext(RouterContext.Location).Path

compiles with no warning and throws `NullReferenceException` when no router is mounted. The repository's own tests read the context that way.

This is the mechanism that made the `UseLocation()` annotation defect possible in the first place (#399). That hook is honest now — it declares `RouterLocation?` — but the public read-path underneath it still is not, so the same trap is one call away, in the same file, for every context whose default is null.

The decision is what `UseContext` should promise. Returning `T?` is honest but pushes a null check onto every consumer including those whose context genuinely has a non-null default. Throwing when a context is absent is React's answer and matches what `useContext` outside a provider does there — but Velvet's contexts are seeded with a default rather than being absent, so "no provider" and "provider with a null default" are the same read, which is the same indistinguishability that decided #399 the other way.

Issue #459, opened 2026-08-06
RouteNavLink with To = "/" renders active when there is no location at all


s4k10503:
`RouteNavLink.Render` computes the current path as `location?.Path ?? string.Empty`, and `Normalize("")` returns `"/"`. So with no router mounted, a `RouteNavLink` whose `To` is `"/"` compares `"/"` against `"/"`, finds them equal, and applies its `ActiveClass`.

A navigation bar rendered before the first navigation — or in any tree with no router above it, such as a preview or a story — shows its home link highlighted as the current page when there is no current page at all.

The empty string is standing in for two different things: "the router says the path is empty" and "there is no router". The first is arguably `"/"`; the second is not a path at all. Whether the fix is an explicit no-location branch or a sentinel that `Normalize` does not fold to `"/"` is the decision this needs.

Found while checking a sentence written for a CHANGELOG entry that claimed the opposite — that `RouteNavLink` relies on the null to render inactive. It does not.

Issue #458, opened 2026-08-06
V.Memoized with deps omitted caches its subtree permanently, and disagrees with Hooks.UseMemo on what omission means


s4k10503:
`V.Memoized(factory)` and `V.MemoizedWithKey(key, factory)` declare `params object?[]? deps` with no deps-less companion overload. Omitting the deps list therefore does not mean "no dependencies" — Roslyn passes the shared `Array.Empty<object>()` instance, `FiberMemoCache.GetOrComputeWithHitInfo` compares it with `ObjectIs.AreEqualDeps`, and `AreEqualDeps` short-circuits on `ReferenceEquals` before comparing anything.

So the comparison sees the same array instance on every render and can never be perturbed. The memoized subtree is cached permanently from its first render, for the element's whole life.

This is the same shape as the `UseBlocker` defect fixed under #398, in a different file.

**It needs a decision rather than a patch.** For a memo, "no deps" is not obviously wrong — React's `useMemo(fn, [])` means exactly "compute once and keep it". But this disagrees with its own closest sibling: `Hooks.UseMemo(factory)` has a deps-less overload that means *recompute every render*. So today the same words mean opposite things two files apart, and neither is documented as the intended one.

Whichever way it is resolved, both spellings should mean the same thing across the memo family, and the one that is not the default needs to be reachable — currently "cache forever" is reachable only by accident and "recompute every render" is not reachable at all.

s4k10503:
Correcting this. I wrote that it is "the same shape as the `UseBlocker` defect" and that the remedy is a decision about which meaning to adopt. The first half is wrong in a way that matters: **the hook-side remedy does not transplant here**, for two reasons.

**A deps-less overload would silently disable memoization across generated code.** `Generators~/src/Velvet.SourceGenerators/MemoizeMethodGenerator.cs:333` emits, for an arity-0 `[MemoizeMethod]`, exactly `global::Velvet.V.Memoized(() => Foo_Impl());` — a one-argument call. Line 226 documents the current binding as the intended semantics ("The arity-0 `V.Memoized` cache returns the same VNode forever"), and a purity diagnostic is built on it. Adding an overload that means "recompute every render" would rebind every generated wrapper and turn memoization off for all of them, with no diagnostic and no diff at the call sites.

[...]

PR #463, opened 2026-08-06
fix(velvet): align three routing hook declarations with what they do

### LINKED ISSUE #398: Hooks.UseBlocker has no single-argument overload, so omitting deps silently freezes the predicate at the mount c...
## What breaks

A consumer writes the idiomatic `Hooks.UseBlocker(attempt => isDirty)` inside a form component. Because `deps` is `params object?[]`, omitting it yields an *empty array*, not null. On the mount render `LastDeps` is committed as `[]`; on every later render `unchanged = deps != null && ObjectIs.AreEqualDeps([], [])` is true, so `NextNeedsReregister` is false and `CommitBlockerSlots` leaves the *original* delegate registered. The blocker therefore evaluates `isDirty` as captured at first render forever: the user edits the form, navigates away, and is not blocked (or is blocked permanently if the first render had it true). No diagnostic fires either — `UseEffectExhaustiveDepsAnalyzer.AnalyzeInvocation` bails with `if (args.Count <= hook.DepsArgIndex) return;` on the comment "The no-deps overloads (factory only) ... are handled by the runtime, not this lint", which is true for UseCallback/UseMemo/UseImperativeHandle (they each ship an explicit no-deps overload) but false for UseBlocker.

## Where

`Packages/com.velvet.core/Runtime/Hooks/Hooks.cs`:460

## Evidence

Hooks.cs:458 `<param name="deps">Dependency array. When null, re-registers on every render.</param>` — unreachable by omission, since `params` supplies `[]`. Hooks.cs:460/475 are the only two UseBlocker signatures (PublicAPI.txt:275-276 confirms), both `params object?[] deps`. Hooks.cs:504-509 `var unchanged = deps != null && ObjectIs.AreEqualDeps(existing.LastDeps, deps); ... existing.NextNeedsReregister = !unchanged; existing.NextRegister = (!unchanged && router != null) ? ... : null;`. FiberHookCommit.cs:125-129 `if (slot.NextNeedsReregister) { slot.Registration?.Dispose(); slot.Registration = slot.NextRegister?.Invoke(); }`. Contrast Hooks.cs:291-295, the UseCallback remark that names exactly this hazard: "The single-argument overload exists so that omitting deps is unambiguous: the `params object?[] deps` overload would otherwise observe an empty array (not `null`) and incorrectly freeze the callback to the first-render closure." The same guard was given to UseMemo (383) and UseImperativeHandle (1304) but not to UseBlocker. Runtime/Routing/Tests/Editor/BlockerTests.cs:471 only ever calls `Hooks.UseBlocker(..., dep)` with an explicit dep, so the omitted-deps path is untested. UseEffectExhaustiveDepsAnalyzer.cs:50-52 is the lint's early-out.

## Independently confirmed

An independent reader tried to refute this and could not:

I tried to refute this and could not; every link in the chain checks out against the source, and I confirmed the two non-obvious ones by compiling repros.

1. Signatures. `Packages/com.velvet.core/Runtime/Hooks/Hooks.cs:460` and `:475` are the only two `UseBlocker` overloads, both `params object?[] deps` (PublicAPI.txt:275-276 agrees). Contrast the siblings, which all ship a deps-less companion: `UseCallback` (PublicAPI.txt:277/278), `UseMemo` (298/299), `UseImperativeHandle` (289/290). The whole `UseEffect`/`UseLayoutEffect`/`UseInsertionEffect` family dodges the hazard a different way — `object?[]? deps = null` as an *optional* parameter (Hooks.cs:851, 868), so omission yields `null` and the effect re-runs every render (React parity). `UseBlocker` is the one deps-taking hook where omission silently means `[]`.

2. Omission really yields an empty array and really compiles. I built a scratch project with the exact two overload shapes: `Hooks.UseBlocker(attempt => isDirty)` resolves unambiguously to the sync overload (the async candidate is eliminated on lambda arity) and arrives with `deps.Length == 0`, `deps != null`. The repo's own UseCallback remark (Hooks.cs:291-295) states this hazard verbatim.

3. The freeze. I mirrored `UseBlockerCore` (Hooks.cs:504-509), `ObjectIs.AreEqualDeps` (ObjectIs.cs:119-145) and `FiberHookCommit.CommitBlockerSlots` (FiberHookCommit.cs:119-133) in a runnable program. Mount registers closure #1 and sets `LastDeps = []`; every later render hits `unchanged = true` (`AreEqualDeps` short-circuits on `ReferenceEquals`, since Roslyn passes the cached `Array.Empty<object>()`, and would return true on length-0 anyway), so `NextNeedsReregister = false` and the commit never disposes/re-registers. Output: with state flipping false→true→true, the registered predicate keeps returning the mount value `false`. `RouteBlockerManager.Register` stores the delegate in a `BlockerEntry` and `CheckAsync` invokes exactly that stored delegate — nothing re-reads a live closure, unlike `UseFrame`, which explicitly routes through a ref (Hooks.cs:898-902) so the latest closure runs without re-subscribing.

4. Nothing catches it. `UseEffectExhaustiveDepsAnalyzer.AnalyzeInvocation` bails at `if (args.Count <= hook.DepsArgIndex) return;` (line ~52) and `UseBlocker`'s descriptor has `depsArgIndex: 1` (DepsHookDescriptor.cs:109-117), so a one-argument call is never analyzed. VEL100 only fires when the array is written out — `UseEffectExhaustiveDepsAnalyzerTests.cs:490` covers `new object[] { }` and reports the missing `isDirty`, which is exactly the failure the omitted form reproduces without a diagnostic. The CHANGELOG entry that added this coverage says the point was that "a captured value missing from their deps went unreported", so the omitted-deps shape defeats the guard's stated purpose.

5. No test and no doc cover it. The only hook-level fixture, `Runtime/Routing/Tests/Editor/BlockerTests.cs:471`, always passes an explicit `dep`. There is no routing guide in `Documentation~/`; the only prose is the XML `<param>` at Hooks.cs:458 describing the null path, and `README.md:141` listing the hook by name. So the behavior is neither documented nor deliberate-looking — and it is not React parity either: React Router's `useBlocker` takes no deps and re-registers whenever the predicate identity changes, i.e. every render for an inline lambda.

Consequence for a consumer: the React-Router-shaped call `Hooks.UseBlocker(attempt => isDirty)` compiles clean, produces no warning, and guards navigation with the mount render's value forever — an unsaved-changes prompt that silently never fires (or, if the mount value was true, one that blocks permanently).

## Correction to the original claim

Two refinements to the report, neither of which weakens it. (a) The documented escape hatch is worse than "unreachable by omission": `UseBlocker` annotates its array non-nullable (`params object?[] deps`), unlike `UseMemo`'s `params object?[]? deps` at Hooks.cs:414, so the "when null, re-registers on every render" path at Hooks.cs:458 is reachable only by passing `null` explicitly and eating a CS8625 nullable warning — I reproduced that warning. (b) The mechanism is one notch more certain than stated: `AreEqualDeps` short-circuits on `ReferenceEquals` before the length check, because Roslyn passes the shared `Array.Empty<object>()` for a zero-argument `params` call, so the comparison cannot be perturbed by dep contents at all. Scope note: this is the sole deps-taking hook with the gap — the effect family uses an optional `= null` parameter rather than `params`, so `UseEffect(() => ...)` correctly means "every render".

## What already exists

Hooks.UseBlocker(Func<NavigationAttempt, bool>, params object?[] deps) and Hooks.UseBlocker(Func<NavigationAttempt, CancellationToken, UniTask<bool>>, params object?[] deps) — both accept an explicit null deps argument that re-registers the fresh predicate every render (Hooks.UseBlockerCore, FiberHookCommit.CommitBlockerSlots). The deps-less overloads that DO exist elsewhere: Hooks.UseCallback<T>(T) and Hooks.UseImperativeHandle<THandle>(Ref<THandle>, Func<THandle>).

PublicAPI.txt lists exactly two UseBlocker members (lines 275-276), both `(delegate, System.Object[])`. Runtime/Hooks/Hooks.cs:460 and :475 declare them as `params object?[] deps`, so there is genuinely no deps-less overload — the claim's title is literally true, and so is its rationale: with `params`, `UseBlocker(pred)` binds an empty (not null) array, and `UseBlockerCore` (Hooks.cs:504) computes `unchanged = deps != null && ObjectIs.AreEqualDeps(existing.LastDeps, deps)`, which is true for two empty arrays (ObjectIs.AreEqualDeps returns true on equal length with no elements, and the compiler-cached empty array even short-circuits on ReferenceEquals). So NextNeedsReregister stays false and the mount render's predicate closure is never replaced — the freeze the claim describes. What the claim overstates is that the capability is unreachable: passing deps explicitly as null is a supported, documented path — the XML doc on both overloads says "Dependency array. When null, re-registers on every render", and the `deps != null &&` guard plus FiberHookCommit.CommitBlockerSlots (Dispose then re-Invoke NextRegister, then LastDeps = NextDeps) make `UseBlocker(pred, null)` re-register the fresh closure every render. The gap is therefore the ergonomic/safety overload, not the behavior. This is a real inconsistency with the codebase's own stated convention: the remark on Hooks.UseCallback<T>(T) (Hooks.cs:291-295) says the single-argument overload "exists so that omitting deps is unambiguous: the `params object?[] deps` overload would otherwise observe an empty array (not null) and incorrectly freeze the callback to the first-render closure" — exactly the hazard UseBlocker still has. UseImperativeHandle carries the same deps-less pairing; UseEffect/UseLayoutEffect/UseInsertionEffect avoid it differently, with `object?[]? deps = null` optional parameters (Hooks.cs:851, :868). The only test call site, Runtime/Routing/Tests/Editor/BlockerTests.cs:471, always passes a dep, so nothing exercises either the omitted-deps freeze or the null-deps path.

### LINKED ISSUE #399: Hooks.UseLocation() is annotated non-nullable but returns null when no router is mounted
## What breaks

`var path = Hooks.UseLocation().Path;` compiles with no nullable warning and throws NullReferenceException in any component rendered outside a Router, or before the router's first navigation has published a location. Hooks.cs is `#nullable enable` (line 2) and its routing siblings are annotated correctly — `UseMatch` returns `RouteMatch?` (673), `UseRouteError` returns `Exception?` (729), `UseOutletContext<T>` returns `T?` (701) — so a consumer reasonably reads the bare `RouterLocation` as a guarantee.

## Where

`Packages/com.velvet.core/Runtime/Hooks/Hooks.cs`:535

## Evidence

Hooks.cs:533-539: `/// Reads <see cref="RouterContext.Location"/>; returns null when no router is mounted.` над `public static RouterLocation UseLocation() { _ = Resolve("UseLocation"); return UseContext(RouterContext.Location); }`. RouterContext.cs:13-14 seeds the default with null: `ComponentContext<RouterLocation>.Create(null)`, and Hooks.UseContext (248) returns `context.DefaultValue!` when no Provider is on the stack. Velvet's own code treats the result as nullable: Hooks.cs:550 `return location?.Params ?? EmptyParams;`, Hooks.cs:678 `RouteQuery.StripQuery(location?.Path)`, and Runtime/Routing/RouteLink.cs:52-53 (a `#nullable enable` file) `var location = Hooks.UseLocation(); var currentPath = location?.Path ?? string.Empty;`. NavigationState.Location, by contrast, is correctly declared `RouterLocation?` (Runtime/Routing/RouterState.cs:94).

## Independently confirmed

An independent reader tried to refute this and could not:

I tried to refute this and could not. Every link verified independently:

1. Nullable context is real and load-bearing: Packages/com.velvet.core/Runtime/csc.rsp contains `-nullable:enable` for the whole runtime assembly, and Hooks.cs:1-2 restates it ("Full nullable checking: hook public surface ... part of the shipped API contract"). So `public static RouterLocation UseLocation()` (Hooks.cs:535) is a non-null promise emitted as [Nullable] metadata to consumers.

2. Found the exact laundering point the report only implied: ComponentContext<T>.DefaultValue is declared `T?` (Runtime/Component/ComponentContext.cs:13), and Hooks.UseContext returns `context.DefaultValue!` (Hooks.cs:248) — the `!` converts a null default into a non-null-annotated T. RouterContext.Location is seeded `Create(null)` (RouterContext.cs:13-14) and RouterLocation is a sealed class (RouterState.cs:60). So neither the definition nor a consumer's `Hooks.UseLocation().Path` yields any diagnostic, and the latter NREs.

3. The null path is supported, not misuse: Velvet's own RouteNavLink.Render (Runtime/Routing/RouteLink.cs:52-53, a #nullable enable file) writes `location?.Path ?? string.Empty` so NavLink renders inactive with no location above it. RouterHooksTests.Given_NoRouterLocationProvider_When_Rendered_Then_LocationIsNull pins the null return, and RoutingHooksTests stores the result in `RouterLocation?` and needs `Capture.Location!.Path`.

4. Reachability confirmed: there is no built-in Router component — the app root provides RouterContext.Location itself (Samples~/StarterApp/StarterApp.cs:56-69 seeds it from `UseState(router.CurrentLocation)`, and Router.CurrentLocation is `RouterLocation?`, null until the first navigation).

5. Nothing guards it: PublicAPI.txt:296-314 erases nullability (it lists `UseMatch(): Velvet.RouteMatch` though the source returns `RouteMatch?`), so no drift test would catch the mismatch, and the sibling hooks UseMatch/UseRouteError/UseOutletContext are annotated nullable, making UseLocation the outlier.

React parity does not rescue it: React Router types useLocation() non-nullable precisely because it throws an invariant outside a router. Velvet neither throws nor annotates nullable, so the current state matches neither React nor the method's own XML doc one line above the signature.

## Correction to the original claim

The finding stands, but two details need trimming and one reframing. (a) Route elements rendered through an Outlet can never observe null: FiberNodePatcher.ResolveOutletMatch (line 1578) returns false when `location?.Matches == null`, so nothing renders — the "before the first navigation" case only bites a component placed as a sibling of the Outlet inside the Provider, or one mounted outside the router root entirely. (b) The cited RouteLink.cs:52-53 sits in RouteNavLink.Render, not RouteLink.Render (same file). (c) Severity: this is a compile-time API-contract defect, not a runtime logic bug — the null return is deliberate, documented at Hooks.cs:533, and covered by RouterHooksTests. Because React Router's non-nullable useLocation() type is backed by an invariant throw, the fix has two parity-valid shapes: annotate `RouterLocation?` (matching UseMatch/UseRouteError/UseOutletContext, and what Velvet's own NavLink and tests already assume), or throw when no location context is above — the latter would require changing NavLink, which currently relies on the null.

## Absence confirmed

An independent check against `PublicAPI.txt` and the code found no equivalent under any name:

Verified in source. `Packages/com.velvet.core/Runtime/Hooks/Hooks.cs` opens with `#nullable enable` (line 2), and at line 535 declares `public static RouterLocation UseLocation()` — a non-nullable annotation — whose body is `return UseContext(RouterContext.Location);`. The value can genuinely be null: `RouterContext.Location` (Runtime/Routing/RouterContext.cs) is `ComponentContext<RouterLocation>.Create(null)` — a null default in a nullable-oblivious file — and `Hooks.UseContext<T>` (line 242-250) hands back `context.DefaultValue!`, the null-forgiving `!` being exactly what suppresses the warning. `RouterLocation` is `public sealed class` (Runtime/Routing/RouterState.cs:60), so null is representable.

Three independent corroborations that null is the real behavior, not a hypothetical: (1) the hook's own XML doc says "returns null when no router is mounted"; (2) `Runtime/Hooks/Tests/Editor/RouterHooksTests.cs` has `Given_NoRouterLocationProvider_When_Rendered_Then_LocationIsNull`, asserting `LastLocation, Is.Null`; (3) the only in-repo consumer, `Runtime/Routing/RouteLink.cs:53`, defensively writes `location?.Path ?? string.Empty` — a null-conditional the annotation says is unnecessary.

The mismatch is specific to this one hook: every sibling router hook that can yield nothing is correctly annotated — `UseMatch(string)` returns `RouteMatch?` (line 673), `UseOutletContext<T>()` returns `T?` (701), `UseLoaderData<T>()` returns `T?` (713), `UseRouteError()` returns `Exception?` (729) — and `UseParams()` sidesteps null entirely by returning `location?.Params ?? EmptyParams`. So the codebase does annotate deliberately; `UseLocation` is the outlier. `PublicAPI.txt:296` records `method Velvet.Hooks.UseLocation(): Velvet.RouterLocation`, but that rendering carries no nullability information either way, so it neither confirms nor refutes on its own.

Nothing in the package supplies the missing annotation under another name (no `[return: MaybeNull]`, no partial declaration — `UseLocation` occurs only in Hooks.cs, PublicAPI.txt, the two test fixtures and RouteLink.cs). The claim holds as stated.

### LINKED ISSUE #400: ISearchParams.Get is annotated non-nullable while its only implementation returns string? and documents null
## What breaks

`Hooks.UseSearchParams()` hands back an `ISearchParams`, so consumers code against the interface. `var (sp, _) = Hooks.UseSearchParams(); var id = sp.Get("id"); if (id.Length > 0) ...` gets no nullable warning from the interface annotation and NREs for any query string that lacks `id` — the interface's own doc comment on the next line says the method returns null in exactly that case. The concrete class is annotated correctly, so the mismatch is invisible unless the caller happens to hold a `SearchParams` rather than the interface the hook returns.

## Where

`Packages/com.velvet.core/Runtime/Routing/SearchParams.cs`:66

## Evidence

SearchParams.cs:65-66 `/// <summary>Returns the first value for a key, or <c>null</c> when the key is absent.</summary>` over `string Get(string key);` (no `?`), versus SearchParams.cs:111-113 `/// <inheritdoc />` over `public string? Get(string key) => key != null && _values.TryGetValue(key, out var list) && list.Count > 0 ? list[0] : null;`. The file carries no `#nullable` directive, but `Runtime/csc.rsp` sets `-nullable:enable` for the whole Velvet assembly (and the file's own use of `string?` only compiles in an enabled context), so both annotations are emitted into metadata. This asymmetry should also be producing a CS8766 on the implementing member, which suggests it was never intended.

## Independently confirmed

An independent reader tried to refute this and could not:

Independently confirmed on every point. (1) SearchParams.cs:66 declares `string Get(string key);` on ISearchParams with the doc comment on line 65 saying it returns null when the key is absent; line 112 implements it as `public string? Get(string key)`. (2) Runtime/csc.rsp carries `-nullable:enable` and sits beside Runtime/Velvet.asmdef, so the whole assembly is in an enabled context and both annotations are emitted. (3) The CS8766 is not hypothetical: it is present in this repo's actual build output — Library/Bee/CachedNodeOutput/*_out and ~/Library/Logs/Unity/Editor.log both contain "SearchParams.cs(112,24): warning CS8766: Nullability of reference types in return type of 'string? SearchParams.Get(string key)' doesn't match implicitly implemented member 'string ISearchParams.Get(string key)'". (4) I reproduced the consumer effect in a minimal netstandard2.1 project with Nullable=enable using the same shapes: the implementation gets CS8766, and a consumer calling `p.Get("id").Length` through the interface gets NO CS8602 at all — the misleading annotation is the one consumers see. (5) Hooks.cs:647 returns `(ISearchParams, SearchParamsSetter)`, so the interface is exactly what callers hold; the path is reachable. (6) No guard: no [MaybeNull], no #nullable directive, no nullability convention test; RoutingHooksTests only asserts present-key values (Get("q") == "velvet"), never the absent-key null case. (7) Not deliberate: git log on the file shows commit e41904e "fix(velvet): align nullable contracts and eliminate CS86xx warnings" — the project explicitly intended to remove CS86xx, and this one survived. It is unrelated to React semantics.

## Correction to the original claim

Two refinements, neither of which overturns the finding. (a) The mismatch is not entirely invisible inside Velvet: CS8766 already fires in Velvet's own compile and is simply being tolerated (the current Editor log shows it 4 times, plus a second unrelated CS8766 at Runtime/Component/V.cs for CastReadOnlyList's indexer — a sibling instance of the same annotation drift). It is hidden from consumers, not from the package build. (b) A downstream consumer is only misled if their own assembly compiles with nullable enabled; Unity assemblies default to disabled, where no annotation is consulted either way. So the blast radius is nullable-enabled consumer assemblies rather than all consumers. The fix is to annotate the interface member as `string? Get(string key);`, which also clears the existing CS8766.

s4k10503:
Three routing hooks declared one thing and did another.

**`UseBlocker` had no deps-less overload.** Because the deps list is `params`, omitting it does not mean "no dependencies" — the compiler passes the shared empty array, and `ObjectIs.AreEqualDeps` short-circuits on `ReferenceEquals` before comparing anything, so the comparison sees the same instance every render and can never be perturbed. The predicate froze at the mount closure. A blocker guarding an unsaved-changes form therefore kept answering from the state the form had when it mounted: never blocking, or — if the mount value happened to be true — blocking permanently. Both silently.

`UseBlocker` now has deps-less overloads for the sync and async forms, meaning "re-register every render" as `UseCallback` and `UseMemo` already do. The deps-taking overloads widen to `params object?[]? deps`, which is `UseMemo`'s annotation; note that only the annotation matches, since `UseBlocker` guards on `deps != null` and re-registers where `UseMemo` compares a null against a null and freezes.

**`UseLocation()` was annotated non-nullable and returns null.** The annotation was the side that was wrong. React Router's `useLocation` throws and is typed non-null because of that invariant, but its invariant tests the router *context object*, and once a Router is mounted its location is always real. Velvet has no built-in Router component: the location context *is* the location, seeded with a null default, so "no router above me" and "router mounted, first location not yet published" are the same observation — both reachable, the second in the starter sample. Throwing would turn a legitimate pre-navigation render into a crash. Every sibling hook is already annotated nullable.

**`ISearchParams.Get` was annotated non-nullable** while its only implementation returns null and documents null. This one was raising CS8766 in the package's own build.

Nullability produces no diff in `PublicAPI.txt`, which renders return types without their annotations — that is why all three could ship, and it is filed separately (#457). Two probe tests pin these members by reading the emitted metadata; they are regression pins for what shipped wrong, not a pattern to propagate, since each one covers only a member someone already suspected.

Four defects found while establishing the above are filed rather than fixed here: `Hooks.UseContext<T>` still promises non-null while returning a null default, so reading a router context off it throws with no warning (#461); an explicit `null` deps list freezes `UseCallback` and `UseMemo` but re-runs `UseBlocker` and `UseEffect` (#462); `V.Memoized` with deps omitted caches its subtree permanently, and the remedy used here does not transfer to it (#458); and `RouteNavLink` with `To = "/"` renders active when there is no location (#459).

No `Documentation~` impact: there is no routing guide, and no guide, index entry or migration section mentions `UseBlocker`, `UseLocation`, `RouterLocation` or `SearchParams`.

Closes #398
Closes #399
Closes #400

### DIFF
--- a/Packages/com.velvet.core/CHANGELOG.md
+++ b/Packages/com.velvet.core/CHANGELOG.md
@@ -66,6 +66,41 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   React does — the success state is dispatched after the handler runs, so a handler that throws never
   reaches it.
 
+- `Hooks.UseBlocker` now has a single-argument overload, and calling it without a dependency array
+  re-registers the predicate on every render. Because deps were declared `params`, omitting them used
+  to bind an *empty* array rather than the null that means "re-register every render", and an empty
+  array compares equal to itself on every subsequent render. The blocker therefore kept the predicate
+  closure the mount render created, and answered forever with the state that render had captured: the
+  idiomatic `Hooks.UseBlocker(attempt => isDirty)` never fired an unsaved-changes prompt, or, if the
+  first render had `isDirty` true, blocked every departure permanently. Nothing reported it — the VEL100
+  exhaustive-deps analyzer skips a call with no deps argument, correctly for the sibling hooks whose
+  omitted form is already safe. `UseCallback`, `UseMemo` and `UseImperativeHandle` carry the same
+  deps-less overload for the same reason, and `UseEffect`, `UseLayoutEffect` and `UseInsertionEffect`
+  reach the same place through an optional `deps = null` parameter. The deps-taking overloads now also
+  accept a null array without a nullable warning; that is `UseMemo`'s annotation, not its behaviour —
+  `UseBlocker` guards on `deps != null` and re-registers, where `UseMemo` and `UseCallback` compare a
+  null against a null, find them equal and freeze.
+
+- `Hooks.UseLocation()` is declared `RouterLocation?`. Its documentation already said it returns null
+  with no router mounted, and Velvet's own `RouteNavLink` already reads it that way, but the
+  declaration promised non-null, so `Hooks.UseLocation().Path` compiled without a warning in a
+  nullable-enabled assembly and threw `NullReferenceException`. The documentation now also names a
+  second case it had left out: a mounted router returns null until it publishes its first location,
+  which is the state `Samples~/StarterApp` starts in, seeding the location context from
+  `Router.CurrentLocation` before the first navigation. The sibling router hooks `UseMatch`,
+  `UseOutletContext`, `UseLoaderData` and `UseRouteError` were already annotated nullable.
+
+  The read path underneath is unchanged and still hands a null past a non-null declaration:
+  `Hooks.UseContext<T>` returns `T` while `ComponentContext<T>.DefaultValue` is `T?`, so
+  `Hooks.UseContext(RouterContext.Location).Path` still compiles clean and throws with no router
+  above it.
+
+- `ISearchParams.Get` is declared `string?`, matching the null it documents for an absent key and the
+  `SearchParams` implementation that returns it. `Hooks.UseSearchParams()` hands back the interface,
+  so the interface declaration is the one a consumer reads: `searchParams.Get("id").Length` compiled
+  without a warning and threw for any query string lacking the key. The mismatch was also raising
+  CS8766 in the package's own build.
+
 ## [2.0.0] - 2026-08-02
 
 ### Highlights
--- a/Packages/com.velvet.core/Generators~/tests/Velvet.SourceGenerators.Tests/GeneratorTestHelper.cs
+++ b/Packages/com.velvet.core/Generators~/tests/Velvet.SourceGenerators.Tests/GeneratorTestHelper.cs
@@ -113,7 +113,12 @@ namespace Velvet
         public static T UseMemo<T>(global::System.Func<T> factory) => factory();
         public static T UseMemo<T>(global::System.Func<T> factory, params object[] deps) => factory();
         public static global::Velvet.RouteBlockerState UseBlocker(
+            global::System.Func<global::Velvet.NavigationAttempt, bool> shouldBlock) => null;
+        public static global::Velvet.RouteBlockerState UseBlocker(
             global::System.Func<global::Velvet.NavigationAttempt, bool> shouldBlock, params object[] deps) => null;
+        public static global::Velvet.RouteBlockerState UseBlocker(
+            global::System.Func<global::Velvet.NavigationAttempt, global::System.Threading.CancellationToken,
+                global::Cysharp.Threading.Tasks.UniTask<bool>> shouldBlock) => null;
         public static global::Velvet.RouteBlockerState UseBlocker(
             global::System.Func<global::Velvet.NavigationAttempt, global::System.Threading.CancellationToken,
                 global::Cysharp.Threading.Tasks.UniTask<bool>> shouldBlock, params object[] deps) => null;
--- a/Packages/com.velvet.core/PublicAPI.txt
+++ b/Packages/com.velvet.core/PublicAPI.txt
@@ -272,7 +272,9 @@ method Velvet.Hooks.ForwardedRef`1(): Velvet.Ref`1[T]
 method Velvet.Hooks.StoreMemoizedVNode(System.Int32, System.Object[], Velvet.VNode): System.Void
 method Velvet.Hooks.TryGetMemoizedVNode(System.Object[], System.Int32&, Velvet.VNode&): System.Boolean
 method Velvet.Hooks.UseAnimationSequence(System.Collections.Generic.IReadOnlyList`1[Velvet.AnimationSequenceStep], System.Boolean, System.Boolean, System.Object[]): System.ValueTuple`2[Velvet.AnimationSequenceState, Velvet.AnimationSequenceControls]
+method Velvet.Hooks.UseBlocker(System.Func`2[Velvet.NavigationAttempt, System.Boolean]): Velvet.RouteBlockerState
 method Velvet.Hooks.UseBlocker(System.Func`2[Velvet.NavigationAttempt, System.Boolean], System.Object[]): Velvet.RouteBlockerState
+method Velvet.Hooks.UseBlocker(System.Func`3[Velvet.NavigationAttempt, System.Threading.CancellationToken, Cysharp.Threading.Tasks.UniTask`1[System.Boolean]]): Velvet.RouteBlockerState
 method Velvet.Hooks.UseBlocker(System.Func`3[Velvet.NavigationAttempt, System.Threading.CancellationToken, Cysharp.Threading.Tasks.UniTask`1[System.Boolean]], System.Object[]): Velvet.RouteBlockerState
 method Velvet.Hooks.UseCallback`1(T): T
 method Velvet.Hooks.UseCallback`1(T, System.Object[]): T
--- a/Packages/com.velvet.core/Runtime/Hooks/Hooks.cs
+++ b/Packages/com.velvet.core/Runtime/Hooks/Hooks.cs
@@ -451,13 +451,33 @@ namespace Velvet
         #region UseBlocker
 
         /// <summary>
-        /// Conditionally blocks navigation departures (synchronous variant).
+        /// Conditionally blocks navigation departures (synchronous variant), re-registering the predicate on
+        /// every render so it answers with the state the latest render captured.
+        /// Must be used inside Render() only.
+        /// </summary>
+        /// <remarks>
+        /// The single-argument overload exists so that omitting deps is unambiguous — see
+        /// <see cref="UseCallback{T}(T)"/> for the hazard it avoids.
+        /// </remarks>
+        /// <param name="shouldBlock">Predicate delegate; returning true blocks the departure.</param>
+        /// <returns>The shared <see cref="RouteBlockerState"/> handle for inspecting / resolving the pending departure.</returns>
+        public static RouteBlockerState UseBlocker(Func<NavigationAttempt, bool> shouldBlock)
+        {
+            if (shouldBlock == null) throw new ArgumentNullException(nameof(shouldBlock));
+            return UseBlockerCore(
+                (router, state) => router.RouteBlockerManager.Register(shouldBlock, state),
+                null);
+        }
+
+        /// <summary>
+        /// Conditionally blocks navigation departures (synchronous variant), re-registering the predicate
+        /// only when a dependency changes.
         /// Must be used inside Render() only.
         /// </summary>
         /// <param name="shouldBlock">Predicate delegate; returning true blocks the departure.</param>
         /// <param name="deps">Dependency array. When null, re-registers on every render.</param>
         /// <returns>The shared <see cref="RouteBlockerState"/> handle for inspecting / resolving the pending departure.</returns>
-        public static RouteBlockerState UseBlocker(Func<NavigationAttempt, bool> shouldBlock, params object?[] deps)
+        public static RouteBlockerState UseBlocker(Func<NavigationAttempt, bool> shouldBlock, params object?[]? deps)
         {
             if (shouldBlock == null) throw new ArgumentNullException(nameof(shouldBlock));
             return UseBlockerCore(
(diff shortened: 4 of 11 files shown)


s4k10503:
CI on this branch is failing on the GitHub Actions incident that began at 15:22 UTC (status page: Actions in major outage, impact critical, still investigating at 16:33 UTC). Every failed job dies in `Set up job` with `Failed to resolve action download info. Error: Service Unavailable` — no workflow step has run, and the aggregate check reports its dependencies as `abandoned` rather than failed.

PlayMode did complete and passed. The rest will be re-run once the incident clears; nothing here needs a code change.

s4k10503:
CI on this branch is blocked by the GitHub Actions incident, not by anything in the change. The status page at 21:30 UTC states that webhook triggers are throttled and "many push and pull request events are not triggering new workflow runs", which is why the checks list is empty.

The tests do pass on this branch's current head. Both workflows completed `success` against `9c04db8` — the same commit this PR points at, with `main` merged in — started via `workflow_dispatch` since that path still works. Those runs cannot satisfy the required checks, because a required check is only settled by a run the `pull_request` event started.

Nothing here needs a code change; it needs the throttle lifted.

Issue #456, opened 2026-08-06
Generators build is not reproducible, so the committed DLLs cannot show whether they match their sources


s4k10503:
> **Corrected after measurement.** The original report below reads the symptom as a non-deterministic build regenerating the MVID on every compile. That is not what was happening, and the comment thread's later diagnosis — that the build "reproduces per environment but not across them" — named a cause its own data could not support, because both of its comparisons were taken at the same commit and only ever varied the path. The measurement is in the next section. The original text is kept unchanged beneath it.

## What was actually happening

Two independent inputs, and the first masked the second.

**1. The commit id.** The SDK queries git and appends `HEAD`'s id to the assembly's informational version.

| | differing bytes |
|---|---|
| two clean builds, one working tree, one commit | **0** |
| one commit, two different checkout paths | **0** |
| **identical sources, two different commits** | **143** |
| a build at `HEAD` vs the committed pair | 144 |

Repeat builds were already byte-identical, and so were builds from two different paths, so neither the MVID nor an embedded path was varying on its own. Read straight out of the two binaries at the same offset:

    fresh build (at 89042a0):  0.1.0+89042a053e8ad7ae408117cbc66c78ccc40ee268
    committed DLL:             0.1.0+acbd80dca9a21e4d388dd1c3e402042b390ecf31

`acbd80d` is twelve commits older than `5c2116f`, the commit that deployed that pair. That is structural rather than an accident: `build.py` necessarily runs before the commit carrying its output, so a deployed assembly can never name the commit it ships in. It also explains, with nothing left over, why a rebuild-and-diff guard failed on the runner — on a pull request the runner's `HEAD` is never the commit a committed DLL was built at.

**2. The .NET runtime the compiler executed on.** With the commit removed, a macOS build and the Linux CI build still differed by 72 bytes, in exactly five runs — PE header timestamp (offset 136), MVID (86288), debug directory timestamp (96832), RSDS PDB GUID (96916) and PDB checksum (97029). Every other byte of both assemblies was identical across both the operating system and the architecture.

A portable PDB records the compilation options, and comparing them names the input: `compiler-version` is the same on both, `runtime-version` is not (`10.0.3-servicing…` against `10.0.11-servicing…`). `global.json` pins the SDK, not the runtime the SDK's tools run on. The assembly carries the PDB's checksum, so a PDB that moves moves the assembly.

## What closes it

[...]

PR #552, opened 2026-08-07
ci(velvet): fail when the committed generator DLLs do not rebuild from their sources

s4k10503:
CLAUDE.md makes redeploying the DLLs part of finishing a generator change, and the committed binaries are what a consumer's compile loads. Nothing checked that they correspond to the sources beside them — and the pair in the tree did not.

**The reproducibility this was thought to need is already there.** The issue reads the symptom as a non-deterministic build regenerating the MVID on every compile. Measured, on one machine in one session:

| | differing bytes |
|---|---|
| two consecutive builds | **0** |
| a build in a worktree vs a build in the original checkout | **0** |
| a build vs the **committed** DLL | **146** |

Neither the MVID nor an embedded path is varying. What differs is the binary in the tree — with the last change to `Generators~/src` and the last deployment of `Runtime/Plugins` sitting in the same commit (`5c2116f`), nothing touching the sources since, and the installed SDK matching `global.json`'s `10.0.103` pin with `rollForward: disable`.

Same sources, same pinned SDK, deterministic compiler, different bytes. Something outside those three decided the committed output, and the binaries could not answer the one question they exist to answer.

**So the guard is worth more than the reproducibility work, and costs one step.** A dirty tree after a build means the tree is stale, not that the build is noisy. It runs on the job that already builds these projects and needs no licence.

Red on the tree as it stood, green after redeploying:

    the committed generator DLLs do not rebuild from the sources beside them
     .../Analyzers/Velvet.SourceGenerators.CodeFixes.dll | Bin 14848 -> 14848 bytes
     .../Plugins/Generators/Velvet.SourceGenerators.dll  | Bin 99840 -> 99840 bytes

Verified with the redeployed DLLs in place: generators 420/420, EditMode 3734/3734, PlayMode 161/161.

What this does not close is the rest of #456 — a build still leaves two binaries in the working tree, so restoring them by path stays a required step for anyone who runs it to check something.


s4k10503:
Withdrawn. The step fails on the runner: the build reproduces per environment but not across them, so a rebuild-and-diff guard would fail every pull request. Measurement and the two viable directions are recorded on #456.

PR #577, opened 2026-08-09
tooling(velvet): refuse a pull-request body that says nothing about where the change came from

s4k10503:
No issue: the guard this branch adds was measured against the pull requests this repository actually
opens, and half of it did not hold.

## What the guard is now

`.claude/hooks/refuse/pr_body_of_another_branch.py` refuses a `gh pr create` whose body says nothing
about where the change came from. An answer is a closing or referring keyword — Closes, Fixes,
Resolves, Refs and their inflections — against the number right after it, or the issue's own URL, or
a `No issue: <reason>` line.

It reads the description that will be posted, so a body file it cannot read is refused rather than
skipped: an absent one, an unexpanded path, stdin, a relative path in a command that also changes
directory, and one the filesystem will not hand over. An inline `--body` is the description rather
than a name for it, so it is searched as it stands — its backticks and `$` cost nothing — and an
expansion in it is refused only where the search came back empty.

Only `gh pr create` is asked. `gh pr edit --body-file` is how an answer gets added after the fact, so
a guard that claimed it would refuse the remedy it hands out.

CONTRIBUTING.md owns the rule and now states it in these terms.

## What it deliberately does not do

**It does not date the body file against the branch.** That half is deleted, not narrowed.

Posed the case it existed for — a body file stamped at the moment one pull request here was opened,
against the branch of the next one opened, whose first commit lands sixteen seconds later — it
allowed it. The window was fifteen minutes wide and the comparison one-sided, so a leftover stamped
after the branch's first commit was allowed with no bound at all. It would not have caught the
occasion it was written for.

No width of window fixes that. A `PreToolUse` hook runs before the command, so where the body is
written by the same command that posts it — a heredoc into the path, then `gh pr create` reading it —
the mtime the guard reads belongs to whatever was already at that path. That is the leftover itself,
and the file it would date is the one the command is about to overwrite. A narrower window refuses
correct bodies; a wider one admits more leftovers; neither is reading the description that will be
posted. A guard that cannot catch its founding incident reads as coverage, which is worse than none.

**It does not require an issue.** A tooling or release change legitimately closes nothing, and the
`No issue:` line answers for it in one line. Measured over the last sixty merged pull requests here:
fifty carry a keyword against a number, one carries a `No issue:` line, two carry only a number
mentioned in passing, and seven carry nothing. So the rule leaves fifty of sixty untouched and asks
one line of the rest — including the two that mention a number in prose, which closes nothing on
merge and is the half of the accident that hurt.

**It does not separate a flag from another option's value.** `-t -Fx` reads a body file of `x`.
Separating them needs a mirror of which of gh's options take a value, and an unpinned mirror drifts.

## Collateral refusals removed

Each was measured against the guard as it stood on this branch, then against the reworked one:

- an inline `--body` containing a backtick was refused as unexpanded, with a remedy that could not be
  followed — the body is the description rather than a name for one, so it is now searched as it
  stands and only an unanswered body that carries an expansion is refused;
- `--dry-run` was refused, and it opens nothing;
- `--head` naming a fork was refused, with the remedy being to pass the head already passed — no head
  is read at all now;
- `--editor`, `--recover`, `--fill-verbose` and the interactive form were passing through an
  exception the guard documented as `--fill`/`--template`; the condition is stated instead of the
  list, since the list was short of what reaches it;
- `builtin cd` was not read as a directory move, so the guard answered about a file in the directory
  the move had left. `builtin` joins `command` in the shared parser's leading words.

## Tests, and how far they reach

`PrBodyProvenanceGuardTests` poses the guard fifty commands and six events, and reads the verdict
each time. A refusal is matched on a phrase only its own message carries **and** on exit 2: a check
that stops firing and a check that fires for another reason both exit 2, and `PreToolUse` runs the
command anyway on every other code, so reading the message alone called a guard that exits 1 a
refusal.

The earlier round claimed every verdict was posed and that was not true. Traced through the guard,
the tables reached **23 of its 29 `return` statements**; they reach all 29 now. The six they missed
were a value attached with `=`, a body flag given as the last token, a segment with no command word,
and the three early returns in `main()` — stdin that is not JSON, an event that is not an object or
not for this tool, and a command that is not text.

**Mutation campaign.** Seventy-five one-line changes to the guard, covering every verdict, every
early return, each accepted spelling, each member of the literal sets and each alternative in the two
patterns. Each is posed the fixture's own tables.

| | caught | survived |
|---|---|---|
| the fixture as this branch last pushed it | 46 | **29** |
| the fixture now | **74** | 1 |

The one survivor is equivalent: `return None` replaced by `break` in `valued()` breaks the only loop
it sits in and reaches the same `return None` two lines below, so no input can tell the two apart.

The campaign is scoped to this fixture alone, which can only make a mutation look more survivable
than it is — `GuardCommandCoverageTests` pins the same guard's recognition of `-Fb.md`,
`--body-file=b.md` and `-bhello`, and would independently have killed several of the twenty-nine.

What the twenty-nine were, and what now kills each:

- **four `NO_ISSUE_LINE` spellings and two widenings.** The line was posed once, unindented,
  lower-case-free, colon-ended and on the first line, so dropping the indent tolerance, the full-stop
  ending, the case-insensitivity or the multi-line anchoring changed nothing — and neither did
  accepting the bare phrase with no reason after it, nor matching it in the middle of a sentence.
  Four bodies now pose the spellings and two pose the widenings, `No issue:` alone and
  "There is no issue: …";
- **`Fixes` and `Resolves`.** Only `Closes` and `Refs` were posed, so half the keyword set was
  deletable. Both are posed now, and so is the adjacency — "Closes the gap … see #123" is a keyword
  and a number that are not one statement, and is refused;
- **the spellings of a body value — eight of the twenty-nine, and the equivalent one is among them.**
  `-F<path>`, `--body-file=<path>` and `-b <text>` appeared only on the accepting side, where losing
  the value reads as "no body" and allows either way. Each now has a refusing row, `-Fs` among them,
  which is the shortest token that carries an attached value and pins the boundary the founding
  accident sat one character from;
- **`--help`, and `gh pr edit`.** Neither was posed; widening the claim from `gh pr create` to
  `gh pr` refused nothing any row asked about;
- **`pushd`/`popd`, a path-spelled command word, an environment-assignment segment, and a move
  carrying past the next segment.** `cd` was the only mover posed, and both relative body paths in
  the table were refused for the move before they were opened, so nothing observed which directory a
  relative path resolves against;
- **four event shapes.** Stdin that is not JSON, JSON that is not an object, a command that is not
  text, and an event carrying no working directory each land in an early return the tables never
  reached. They are a table of their own now, beside the malformed event that was already there and
  a tool outside the declared set;
- **a refusal that exits 1.** The fixture read the reason off the message and returned it whatever
  the exit code was, so a guard that stopped blocking anything still answered "no-origin". Exit 2 is
  now required before the message is read, which is the one of the twenty-nine that made the fixture
  itself wrong rather than incomplete.

`GuardCommandCoverageTests` gained one thing beyond its new body table. Its six per-test
preconditions on how many answers came back are folded into `Disagreements`: a guard that raises on
one command answers none of them, and an `Assume` reported the run that proves it broken as
inconclusive, which the runner does not count as a failure.

## Which of this and #588 merges first

#588 adds a required `HOOK_TOOLS` declaration to every `PreToolUse` hook and a fixture that fails on
a registration without one. This branch registers a new `PreToolUse` hook, so whichever of the two
merged second would have broken the other's fixture on `main`. The hook here therefore declares
`HOOK_TOOLS = {"Bash"}` and gates on it, which makes either order safe.

**Recommend #588 first.** Its fixture then already guards this registration when it lands, and
CONTRIBUTING.md's statement of what `HOOK_TOOLS` is arrives with the check rather than after a hook
that already uses it. Nothing breaks the other way round; it just leaves a declaration undocumented
for as long as the gap lasts.

CONTRIBUTING.md owns the rule and is updated here: `Resolves` joins the keyword list, `--help` joins
`--dry-run` as opening nothing, and the section now says that only `gh pr create` is asked. No
`Documentation~` guide describes this — the guard is contributor tooling, not package behaviour, so
`Packages/com.velvet.core/CHANGELOG.md` gets no entry.

## Suites

EditMode 3781/3781 and PlayMode 161/161, 0 failed and 0 inconclusive in both,
`assert_no_inconclusive.py` clean on each.

### DIFF
--- a/.claude/hooks/lib/shell_commands.py
+++ b/.claude/hooks/lib/shell_commands.py
@@ -15,8 +15,10 @@ import shlex
 SEPARATORS = set(";&|\n")
 
 # Words that may precede the command without changing which command it is. `then`/`do`/`else`
-# because a command inside a conditional or a loop is still that command.
-LEADING_WORDS = {"then", "do", "else", "elif", "!", "time", "command", "nohup", "exec"}
+# because a command inside a conditional or a loop is still that command; `builtin` alongside
+# `command` because a guard reading the word after it saw `builtin cd` as neither a move nor a
+# command word, and answered about a file in the directory the move had left.
+LEADING_WORDS = {"then", "do", "else", "elif", "!", "time", "command", "builtin", "nohup", "exec"}
 ENV_ASSIGNMENT = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*=")
 REDIRECTION = re.compile(r"^\d*(?:>>|>&|<&|>|<)")
 
--- a/.claude/hooks/lib/shell_commands.py
+++ b/.claude/hooks/lib/shell_commands.py
@@ -194,28 +194,11 @@ def leading_program(tokens):
     return index
 
 
-# gh's own options, before the subcommand. Only the value-taking ones need naming: a boolean flag is
-# one token and is stepped over either way. `gh -R owner/repo pr create` is the spelling that made
-# this necessary — every guard reading a gh subcommand walked straight past it.
-GH_GLOBAL_VALUE_FLAGS = {"-R", "--repo"}
-
-
-def past_global_flags(tokens, index, value_flags):
-    """The index of the subcommand, stepping over a program's own options and their values."""
-    while index < len(tokens) and tokens[index].startswith("-"):
-        flag, sep, _ = tokens[index].partition("=")
-        index += 1 if sep or flag not in value_flags else 2
-    return index
-
-
 def program_invocations(command, program, words):
     """Operands after `program` followed by `words`, once per segment that runs it.
 
     For programs whose subcommand is a fixed word sequence — `gh issue create`, `gh pr merge`.
     A pattern over the whole command answered yes to the words appearing inside an argument.
-
-    A program's own options come before the subcommand and are stepped over. Matching from the token
-    straight after the program name is what let `gh -R owner/repo pr create` past every guard here.
     """
     found = []
     for segment in command_segments(command):
@@ -223,7 +206,7 @@ def program_invocations(command, program, words):
         index = leading_program(tokens)
         if index >= len(tokens) or os.path.basename(tokens[index]) != program:
             continue
-        index = past_global_flags(tokens, index + 1, GH_GLOBAL_VALUE_FLAGS)
+        index += 1
         if tokens[index:index + len(words)] != list(words):
             continue
         found.append(tokens[index + len(words):])
--- a/.claude/hooks/lib/shell_commands.py
+++ b/.claude/hooks/lib/shell_commands.py
@@ -194,11 +194,28 @@ def leading_program(tokens):
     return index
 
 
+# gh's own options, before the subcommand. Only the value-taking ones need naming: a boolean flag is
+# one token and is stepped over either way. `gh -R owner/repo pr create` is the spelling that made
+# this necessary — every guard reading a gh subcommand walked straight past it.
+GH_GLOBAL_VALUE_FLAGS = {"-R", "--repo"}
+
+
+def past_global_flags(tokens, index, value_flags):
+    """The index of the subcommand, stepping over a program's own options and their values."""
+    while index < len(tokens) and tokens[index].startswith("-"):
+        flag, sep, _ = tokens[index].partition("=")
+        index += 1 if sep or flag not in value_flags else 2
+    return index
+
+
 def program_invocations(command, program, words):
     """Operands after `program` followed by `words`, once per segment that runs it.
 
     For programs whose subcommand is a fixed word sequence — `gh issue create`, `gh pr merge`.
     A pattern over the whole command answered yes to the words appearing inside an argument.
+
+    A program's own options come before the subcommand and are stepped over. Matching from the token
+    straight after the program name is what let `gh -R owner/repo pr create` past every guard here.
     """
     found = []
     for segment in command_segments(command):
@@ -206,7 +223,7 @@ def program_invocations(command, program, words):
         index = leading_program(tokens)
         if index >= len(tokens) or os.path.basename(tokens[index]) != program:
             continue
-        index += 1
+        index = past_global_flags(tokens, index + 1, GH_GLOBAL_VALUE_FLAGS)
         if tokens[index:index + len(words)] != list(words):
             continue
         found.append(tokens[index + len(words):])
--- a/.claude/hooks/refuse/metadata_less_create.py
+++ b/.claude/hooks/refuse/metadata_less_create.py
@@ -13,7 +13,8 @@ asks; this refuses.
 Assignee is required on an issue and not on a pull request: a pull request already records its
 author, so a self-assignment adds nothing.
 
-`--web` hands the fields to the browser form, where they can be set, so it is left alone.
+`--web` hands the fields to the browser form, where they can be set, so it is left alone, and so is
+`--help`, which opens nothing.
 """
 
 import json
@@ -29,6 +30,17 @@ import repository
 # refusal by naming the command it refuses, which is exactly the text that would.
 LABEL_FLAGS = ("--label", "-l")
 ASSIGNEE_FLAGS = ("--assignee", "-a")
+
+
+# --web hands the fields to the browser form, where they can be set; --help opens nothing. Named
+# once so GuardCommandCoverageTests asks the guard rather than restating the rule — the copy in that
+# table listed only --web, and a row added for --help disagreed with the guard that had exempted it.
+EXEMPT_FLAGS = ("--web", "--help", "-h")
+
+
+def exempt(operands):
+    """Whether this invocation creates nothing a label could go on."""
+    return any(token in EXEMPT_FLAGS for token in operands)
 
 
 def carries(operands, flags):
@@ -74,7 +86,7 @@ def main():
     missing = []
     kind = ""
     for kind, operands in creations(command):
-        if "--web" in operands:
+        if exempt(operands):
             continue
         missing = []
         if not carries(operands, LABEL_FLAGS):
--- a/.claude/hooks/refuse/metadata_less_create.py
+++ b/.claude/hooks/refuse/metadata_less_create.py
@@ -13,8 +13,8 @@ asks; this refuses.
 Assignee is required on an issue and not on a pull request: a pull request already records its
 author, so a self-assignment adds nothing.
 
-`--web` hands the fields to the browser form, where they can be set, so it is left alone, and so is
-`--help`, which opens nothing.
+`--web` hands the fields to the browser form, where they can be set, so it is left alone, and so are
+`--help` and its short `-h`, which open nothing.
 """
 
 import json
--- a/.claude/hooks/refuse/pr_body_of_another_branch.py
+++ b/.claude/hooks/refuse/pr_body_of_another_branch.py
@@ -62,9 +62,8 @@ from shell_commands import command_segments, leading_program, program_invocation
 # reads this declaration to check that the registration is still there.
 HOOK_SCOPE = "session"
 
-# The tools this acts on, gated on below rather than restated there: the same fixture compares this
-# set against the matcher that routes tools to it, and a gate spelling the name a second time would
-# drift against both.
+# The tools this acts on, gated on below rather than spelled a second time there: two statements of
+# the same set drift.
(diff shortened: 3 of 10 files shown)


s4k10503:
Round 2 addressed. The campaign was re-run from scratch rather than taken on report, and it came out
larger than the one reported: **75 mutations, 46 caught and 29 survived against the fixture as this
branch last pushed it; 74 caught and 1 survived now.** The single survivor is equivalent — `return
None` replaced by `break` in `valued()` breaks the only loop it sits in and falls to the same
`return None` two lines below, so no input separates them. No other mutation is being called
equivalent.

The mutations cover every verdict, every early return, each accepted spelling, each member of the
literal sets (`MOVERS`, `EXEMPT_FLAGS`, both body-flag tuples) and each alternative in the two
patterns. They were posed the fixture's own tables, parsed out of the `.cs` file so the replica
cannot drift from it; two of them — a refusal dropped to exit 1, and `NO_ISSUE_LINE` widened to
accept the bare phrase — were re-run through the real Unity fixture and reddened it with the
expected text.

On the three findings named:

- **The title was false.** It said the guard refuses a body "that was not written for this branch",
  which is the staleness half that was deleted. It now says what the guard does.
- **"Every verdict is posed" was false.** Traced through the guard, the tables reached 23 of its 29
  `return` statements. They reach all 29 now, and the body states the number rather than the
  universal.
- **The four unpinned `NO_ISSUE_LINE` verdicts.** Six, measured: the indent tolerance, the full-stop
  ending, the case-insensitivity and the multi-line anchoring were all deletable, and so were two
  widenings — accepting the bare phrase with no reason after it, and matching it mid-sentence. All
  six have a body now.

One finding of the fixture's own, not in the list: it read the refusal reason off the message
whatever the exit code was, so a guard dropped to exit 1 — which `PreToolUse` runs the command
through anyway — still answered `no-origin`. Exit 2 is now required before the message is read.
Separately, `GuardCommandCoverageTests` had six `Assume`s on how many answers came back; a guard
that raises answers none, and the run that proves it broken reported inconclusive. Those are folded
into the comparison.

**Merge order with #588.** This branch registers a new `PreToolUse` hook, so whichever of the two
landed second would have broken the other's fixture. The guard here declares `HOOK_TOOLS = {"Bash"}`
and gates on it, which makes either order safe; #588 first is still the better one, so that its
check and CONTRIBUTING.md's statement of what the declaration is arrive together.

EditMode 3781/3781 and PlayMode 161/161, 0 failed and 0 inconclusive in both,
`assert_no_inconclusive.py` clean on each.

s4k10503:
Review round: one wrong claim, one one-directional pin, and four smaller ones.

**The claim about the sibling fixture was wrong.** The commit said this guard "does not break the fixture that pins every PreToolUse hook's tool set against its matcher whichever of the two lands first". It does. That fixture enrols any hook declaring `HOOK_TOOLS` by scanning `.claude/hooks/**/*.py`, and none of its seven payloads was a `gh pr create` — this guard exits 0 with both streams empty under every one of them, with `cwd` an empty directory. On a scratch merge of both branches onto `main`, unmodified:

```
.claude/hooks/refuse/pr_body_of_another_branch.py answers nothing posed as Bash
```

The payload that answers it is added on the other branch. Nothing changes here for it beyond the comment.

The related comment on `HOOK_TOOLS` said "the same fixture compares this set against the matcher that routes tools to it". That is also false at this branch's HEAD — `HookWiringCoverageTests` here reads `HOOK_SCOPE` only, so the set is an unpinned mirror of `settings.json` until the comparison lands. The comment now states the local decision and nothing about a fixture.

**`MOVERS` was pinned in one direction only.** Dropping a member reddens the table; adding one did not. Measured against the whole 51-row command table plus the six event rows, all green under each of: `MOVERS ∪ {"git"}`, `MOVERS ∪ {"echo"}`, `moves_directory` returning `True` unconditionally, and a prefix match. Under `MOVERS ∪ {"git"}` the repository's own way of opening a pull request — a `git push` and a `gh pr create` joined by `&&` — flips from allow to refuse with every row still green. The `BODY=x && …` row does not cover it: it returns early at the environment-assignment segment.

Two negative rows close it, one per command word, since a single row only pins the word it names:

| mutation | table as it stood | with the rows |
| --- | --- | --- |
| `MOVERS ∪ {"git"}` | green | 1 row red |
| `MOVERS ∪ {"echo"}` | green | 1 row red |
| `moves_directory` → `return True` | green | 2 rows red |

**The `resolves.md` fixture masked the keyword it was named for.** It held `Resolved #12` while CONTRIBUTING advertises `Resolves`, so narrowing `resolve[sd]?` to `resolved` left the fixture green — posed directly: with the original body the mutant exits 0, with `Resolves` it exits 2. The same held the other way for `close[sd]?`, `fix(?:e[sd])?` and `refs?`, whose alternatives were posed by nothing. Every spelling of all four keywords now has a row; each narrowing to a single form reddens the rows for the others (2, 2, 2 and 1 respectively).

**The `--template` row read `closes.md`,** so it passed whether or not `--template` was read as a body file at all. Pointed at `silent.md` it dies when `--template` joins `BODY_FILE_FLAGS`.

**Two smaller ones.** The comment over the row where a move holds past an intervening segment stated the opposite of the code — `moved = moved or moves_directory(segment)` is sticky — and nothing in the row needs it, so it is deleted. And the commit message says "five event shapes that land in an early return no command reaches"; there are four. The `cwd` of `12345` row lands in the exception handler and the row with no `cwd` reaches the opener.

**Not fixed.** The `--head someone:feat/x` row still duplicates the plain `closes.md` row against the guard as it stands — no mutation of current code separates them. It is kept as a pin on the removed fork check, which it would catch if that check came back, but it is doing nothing against the code that is there.

Full runs on this branch: EditMode 3781/3781, PlayMode 161/161, `assert_no_inconclusive.py` clean on both. Merged with `tooling/hook-tool-set-drift` onto `main`: EditMode 3792/3792, clean. No CHANGELOG entry — contributor tooling, no user-visible behaviour.

s4k10503:
Pushed 6178872.

**The HIGH is confirmed.** The previous commit's message claimed four mutations "all survived the whole fixture" and that "each now reddens at least one". The prefix match still survived. Measured against all three tables (31 accepted + 28 refused + 6 events), driving each row through the hook exactly as the fixture does:

| mutation of `moves_directory` | before | after |
| --- | --- | --- |
| `MOVERS` with `"git"` added | caught (1 row) | caught (1) |
| `MOVERS` with `"echo"` added | caught (1 row) | caught (1) |
| `MOVERS` with `"cd"` dropped | caught (4 rows) | caught (4) |
| `return True` unconditionally | caught (2 rows) | caught (4) |
| `any(word.startswith(m) for m in MOVERS)` | **survived, 0 disagreements** | caught (1) |
| `any(m.startswith(word) for m in MOVERS)` | **survived, 0 disagreements** | caught (1) |

Two rows close both directions — `cdk` starts with a mover, `push` is the start of one, neither is a mover. Each is RED in Unity under its own loosening and green otherwise:

```
Given_ABodyThatNamesNoOriginOrCannotBeRead_When_TheGuardAnswers_Then_EachRefusalIsItsOwn
  Expected: (30, )
  But was:  (30, cdk deploy && gh pr create --title x --body-file silent.md
    expected [no-origin] got [moved])
```

```
  But was:  (30, push && gh pr create --title x --body-file silent.md
    expected [no-origin] got [moved])
```

**Both LOWs are confirmed and fixed.**

The deletion of "A move holds for the rest of the command, not for the segment after it" was right on the deletion test, but the reason given for it was wrong: read as the ordinary contrastive idiom, the sentence agrees with `moved = moved or moves_directory(segment)`. Stating that plainly rather than quietly, since this is the second consecutive round in which a sentence was reclassified on a misreading rather than on what it said.

The claim that "the repository's own way of opening a pull request runs `git push` in the segment before the create" is supported by nothing in `CONTRIBUTING.md`, `.claude/` or `scripts/` — searched all three. Deleted; the load-bearing half stays.

The pushed message cannot be corrected without a force-push, so the correction is stated in the new commit's message.

EditMode: 3781 passed, 0 failed, 0 inconclusive, 0 skipped; `assert_no_inconclusive.py` clean. No CHANGELOG entry — contributor tooling. No `Documentation~` impact: the guard's behaviour is unchanged this round.

s4k10503:
PlayMode on the pushed head: 161 passed, 0 failed, 0 inconclusive, 0 skipped.

Issue #436, opened 2026-08-06
No finite repeat count or alternate/yoyo direction knob for looping animations


s4k10503:
> **Re-measured against `origin/main` (`89042a0`) on 2026-08-13.** `animate-spin` shipped in #579
> and is no longer part of this issue; the two remaining absences below were re-confirmed at that
> commit. The original body's third absence and the assertion that pinned it are struck through
> rather than deleted, because they are the record of what was true when this was filed.

## What uGUI does

SetLoops(-1, LoopType.Yoyo) or SetLoops(3), and Animator's Loop Time on a clip. This is how every breathing call-to-action, bouncing down-arrow and pulsing objective marker is built, and it is a one-liner.

## What Velvet does

The only class-driven loop is the animate-* family, which recognises exactly five modes — Gradient, Shimmer, Hue, Pulse, Spin — each infinite, each accepting only a duration override (`StyleAnimateClass.cs:12-30`, `136-150`). The general-purpose loop is UseAnimationSequence(loop: true), whose loop flag is a bool (infinite only, Hooks.cs:1027) and which bumps a state version on each step, so the component re-renders on every keyframe (Hooks.cs:1042/1058).

## What an author hits

A bounce, a yoyo, a shake, or "repeat this three times" has no utility and no transition knob. The workaround is a coordinator Motion plus a variant map plus a looping sequence hook that re-renders the component on every keyframe for as long as it is on screen — where uGUI spent one component and no per-frame managed work. UI Toolkit has no @keyframes, but Velvet already runs its own loop ticks for animate-*, so the knobs it does not cover are Velvet's own scope decision.

## Where

/Users/apple/dev/velvet/Packages/com.velvet.core/Runtime/Styling/StyleTransitionConfig.cs:42-173, /Users/apple/dev/velvet/Packages/com.velvet.core/Runtime/Hooks/SequenceWalker.cs:74-97, /Users/apple/dev/velvet/Packages/com.velvet.core/Runtime/Hooks/Hooks.cs:1027

## Confirmed against the codebase

Each of the two remaining absences is Velvet's to close. Read the original framing ("looping animation is a closed set of four modes") as too strong — the capability is reachable, the declarative knob is not.

CONFIRMED ABSENT (searched every plausible surface):

[...]

PR #579, opened 2026-08-09
feat(velvet): add animate-spin

s4k10503:
Closes part of #436 — the `animate-spin` half. The repeat count and the alternate/yoyo direction knob
that issue also asks for are a separate change to `SequenceWalker.Advance`, and are not here.

The namespace was already held open for this: `StyleAnimateClass` claims `animate-*` as a whole, a
comment named `animate-spin` as the future mode, and a test pinned the non-claim. Tailwind settles
every question the utility raises — one turn a second, linear, forever — so nothing here is a
judgement call.

## What it costs to run

The same as the four modes beside it: a phase read off a panel-root tick and one inline write per
frame. Rotation is a value-compared struct, so writing it each frame dirties the element correctly,
without the fresh-list-every-frame that `animate-hue` needs for the filter slot.

Linear rather than eased: the loop restarts at full speed, so an eased turn reads as a stutter at the
wrap.

## Slot ownership

`animate-spin` owns `rotate` while it runs, which is how `animate-hue` owns the filter and
`animate-pulse` owns opacity — a static `rotate-*` is shadowed, and detaching nulls the slot so the
reconciler re-asserts whatever class was under it.

The combination it cannot serve is a Motion rotate channel on the same element: both write the slot
every frame, so the result is whichever wrote last rather than a blend. That is stated in the guide
rather than left for an author to discover.

## Documentation

No guide owned the `animate-*` family — `motion.md` covers `V.Motion` and the timeline hook, and the
utilities appeared only in passing lists. It gains a section: the five modes, their default loops,
the bracketed override, which slot each takes, and the one combination to avoid.

## Verification

- `animate-spin` resolves to a 1s spin; `animate-spin-[2500ms]` to 2.5s; the phase is 90° / 180° /
  360° at a quarter, half and whole of the loop.
- The test that pinned the non-claim now pins it with a name this layer still does not recognise, so
  the open namespace is still held.
- EditMode 3780/3780, `test_release_notes.py` 26/26.


s4k10503:
## Review round: the suspension was wired to the wrong event

All seven findings verified against the code before fixing. One of them was wrong; the rest held, and
the first was worse than reported.

### HIGH 1 — the per-frame suspension, and the `animate-pulse` regression

Confirmed, and it is not narrow. `DeclaredSlots` returns `All` whenever the element carries an inline
`transition-duration`, which is exactly what `StyleAnimationScheduler.ApplyTransitionStyles` writes
and holds for the whole length of a variant swap. Called from the tick, the guard therefore
overwrote the swap's own `transition-property: all` with `none` on every frame — the swap could never
win. `GuardedSlots` maps `Pulse => Opacity`, so this reached elements carrying no `animate-spin` at
all: any shipped `animate-pulse` under a `V.Motion` variant swap or an `AnimatePresence` fade.

The suspension is now taken once at attach, where both Motion drivers take theirs.

RED without the fix (`Given_APulseOverNoDeclaredTransition_When_AVariantSwapRunsUnderIt_Then_TheSwapKeepsItsTransition`):

```
Expected: (<unset>, all)
But was:  (<unset>, )
```

The second element of the tuple is the element's inline `transition-property` after a real
`V.Motion` label flip plus two pulse frames — an empty name is the suspension, `all` is the swap's.

### HIGH 2 — never released

Confirmed. `animate-spin duration-300` patched to `animate-spin transition-colors` left
`transition-property: none` inline until detach. The decision is now re-asked on each patch that
leaves the motion running, through a new `MotionNativeTransitionGuard.SyncSuspension`.

RED (`Given_ASuspendedSpin_When_APatchLeavesNothingTransitioningRotate_Then_TheSuspensionIsHandedBack`):

```
Expected: (Undefined, Null)
But was:  (Undefined, Undefined)
```

### One consequence worth stating, because it shaped the fix

The patch-time decision deliberately does **not** read the inline `transition-duration`. That slot is
also what a variant swap sets, and the swap's `PlayVariantEnter` runs *before*
`ApplyAnimateOnPatch` in `PatchMotion` — so a level-triggered re-ask that read it would have
reintroduced HIGH 1 through the patch instead of the tick. Reading the classes only costs the bracket
form `duration-[400ms]`, which never enters the class list; `motion.md` now says so. A second test,
`Given_APulseSuspendingItsOwnTransition_...`, pins that a patch does not re-assert a suspension the
binding already holds, which is the other half of the same hazard.

### MEDIUM 3 — the "one inline write per frame" claim

The claim was false **as the code stood**, and so were both sentences it was said to contradict:
`MotionNativeTransitionGuard.cs`'s "one inline write per play (never per tick)" and `motion.md`'s
"the way Motion's own drivers do". The right correction was the code, not the prose — all three are
true again now, and the PR body needs no edit.

### MEDIUM 4 — CHANGELOG

`animate-pulse` does change behaviour: it gains a suspension it never had. Entry added under
`## [Unreleased] ### Fixed`, including the cost (the suspension is element-wide while it lasts) and
the two cases where nothing happens.

### LOW 5 / 6 / 7

- The comment carrying the issue number is gone with the call site. What replaced it names the guard
  rather than re-explaining it, and the unpinned engine claim ("on a wrap, travels the long way
  back") is not written anywhere.
- Stale `Hue`/`Pulse` enumerations corrected across `FiberAnimateMotionApplier` and
  `StyleAnimateDriver`, and `RestoreSharedInlineSlot`'s three-way mode test replaced with
  `!IsPanMode(...)` so the next mode does not have to remember to join it.
- The double mount is gone: one tree, two children, via a `MountPair` helper.

### Also found

- The fixture's `ModeCases` / `DefaultDurationCases` tables are documented as one row per recognised
  mode and were a row short of that after this PR. `animate-spin` rows added; the standalone
  `Given_TheSpinToken_...` test they duplicate is removed.
- `Given_ASpinUnderADuration_...` **passed under the broken code** as originally written, because
  `ForcePanelUpdate` ticks the panel scheduler and so runs the animation's first frame. `MountPair`
  now skips the layout pass, which is what makes the test discriminate attach-time from frame-time.

RED for it:

```
Expected: (Null, Undefined, Null)
But was:  (Null, Null, Null)
```

### Suites

EditMode 3787/3787, PlayMode 161/161, both with graphics, both on an otherwise idle machine, zero
inconclusive by `assert_no_inconclusive.py`. `test_release_notes.py` 26/26.

s4k10503:
## Round 2 fixes — pushed as `03da87a`

Both findings reproduced before fixing. Both are now pinned by tests that are RED on `73680b2` and GREEN after.

### HIGH — the suspension did not survive a variant swap

Confirmed. `StyleAnimationScheduler.ClearTransitionStyles` ran at the end of every play and nulled `transition-property` outright, while `SyncSuspension` writes it only on the patch that adds the owner — so `animate-pulse duration-300` on a Motion lost its suspension permanently after one flip, and every frame the pulse wrote afterwards started a native transition instead of landing.

The clear now asks the guard what to leave behind (`MotionNativeTransitionGuard.RestoreAfterForeignWrite`): the suspension while a driver still holds one, the cascade otherwise.

RED on `73680b2`, `Given_APulseSuspendedByADuration_When_AVariantSwapHasRunItsCourse_Then_TheNextPulseFrameIsPaintedAsWritten`:

```
Expected: 0.5f +/- 0.00999999978f
  But was:  0.858796299f
```

The test drives a real flip on a Motion carrying `duration-300 animate-pulse`, plays the swap out to its completion timeout on a stepped panel clock, and then asserts what the panel PAINTS for the next pulse frame. `0.8588` is the engine mid-transition towards the driver's write; `0.5` is the write landing.

### MEDIUM — the release cancelled the swap's own transition

Confirmed, with one correction: the release does null the swap's `transition-property: all`, but that is only visible where the element's classes name a narrower list. On `transition-transform animate-spin` the fall-back is `translate, scale, rotate`, so an opacity swap snapped. With no transition utility present the computed value falls back to UI Toolkit's initial `all` and the swap survives — the finding is real, its blast radius is elements carrying a `transition-*` utility.

Resolved on the guard side rather than by reordering the patcher: the guard now writes and reverts only a value it still holds, and the swap's own teardown restores the suspension afterwards. The ordering constraint is stated in a comment on `HoldsAForeignValue`.

RED on `73680b2`, `Given_ASpinSuspendingTheElement_When_APatchStopsItWhileSwappingAVariant_Then_TheSwapStillTweens`:

```
Expected: less than 0.899999976f
  But was:  1.0f
```

### One more of the same shape, found while fixing

The mirror case was broken too: a patch that STARTS a spin while swapping a variant reached `SyncSuspension` with the swap's `all` in the slot and overwrote it with `none`, so that swap snapped as well. Same fix, same oracle — `Given_AnElementTransitioningTransform_When_APatchStartsASpinWhileSwappingAVariant_Then_TheSwapStillTweens`, RED at `1.0f`.

### Verification

- Full EditMode: 3791 tests, 3791 passed, 0 failed, 0 inconclusive (`assert_no_inconclusive.py` clean). Full PlayMode: 161/161, 0 inconclusive.
- Round 1's `animate-pulse` regression test is unchanged and still discriminating: re-instating round 1's bug (the per-frame `SuspendIfIntercepted` at the top of `ApplyFrame`) on top of this round's code leaves `Given_APulseSuspendingItsOwnTransition_When_AVariantSwapRunsUnderIt_Then_TheSwapKeepsItsTransition` RED — `Expected: (, all) But was: (, )`.
- Arrangement sweep on the three new cases: strip the arranged condition from each (`duration-300` from the first, `transition-transform` from the other two) and all three PASS on the unfixed code, so each arrangement is what makes its case discriminate rather than decoration.

### Notes

- The first implementation compared the slot by list identity and three existing tests caught it: the value read back out of the inline slot is not the list that was assigned. The comparison is by content, and a characterization pin in `MotionNativeTransitionGuardSuspensionTests` fails if that ever changes.
- Docs: `Documentation~/motion.md`'s animate-* section gains a sentence on who owns the slot during a swap. CHANGELOG: no new entry — the suspension itself is unreleased, so the existing `[Unreleased]` `animate-pulse` entry is extended rather than duplicated.

s4k10503:
Rebased onto main. The branch was DIRTY, and a conflicting branch never starts CI at all — so "no checks reported" was not a run in progress, it was a run that could never begin.

One conflict, in `CHANGELOG.md`: this branch's `animate-pulse` entry against the `TabIndex`/`DelegatesFocus` entry that landed on main meanwhile. Both belong under Unreleased ▸ Fixed and both are kept; nothing was dropped and no dated section was touched.

Full EditMode suite on the rebased tree: **3798/3798**, 0 inconclusive, 0 `error CS`. That is the whole suite on the post-rebase code, not the pre-rebase measurement carried forward — the rebase moved five commits over a main that had gained the pooled-widget and focus-defaults work, and the animate-* suspension touches the same styling seams.

Issue #412, opened 2026-08-06
A live registry portal never re-resolves its target: re-registering the id leaves children patching a destroyed element


s4k10503:
## What React does

The container is part of the portal's identity. `reconcileSinglePortal` compares `containerInfo`; when the container object changes, the existing portal fiber is deleted and a new one created, so the children unmount from the old node and mount into the new one on the very next render.

## What Velvet does

`ReconcileKeying.CanPatch` compares `TargetId`, so changing the *string* remounts correctly. But `FiberNodePatcher.ResolvePortalTarget` short-circuits on the recorded target: "A live portal keeps the target its children mounted into: re-registering the id only points FUTURE portals elsewhere." So when the element behind an unchanged id is replaced, the live portal keeps patching into the element it first resolved.

## What a developer hits

A portal target that the UI itself owns — a `"modal-root"` div rendered by a screen that gets torn down and rebuilt on navigation, then re-registered from the new element's `refCallback` — leaves every already-mounted portal writing into the destroyed element. The children vanish from the panel with no warning and no way to recover short of remounting the portal (changing its `targetId`, or keying it so it remounts). React has no equivalent failure because the container is passed by reference and a swap is observed as a change.

## Where

/Users/apple/dev/velvet/Packages/com.velvet.core/Runtime/Reconciler/FiberNodePatcher.cs (ResolvePortalTarget, ~line 816), /Users/apple/dev/velvet/Packages/com.velvet.core/Runtime/Reconciler/ReconcileKeying.cs (CanPatch PortalNode case)

## Confirmed against the codebase

Confirmed by reading the code. There is no other mechanism, under any other name, that makes a live registry portal re-resolve its target.

**What the code does**

`/Users/apple/dev/velvet/Packages/com.velvet.core/Runtime/Reconciler/FiberNodePatcher.cs`, `ResolvePortalTarget` (line 791), registry branch at lines 816-824:

```csharp
describe = newNode.TargetId!;
if (_ctx.PortalState.TryGetValue(placeholder, out var recorded) && recorded.Target != null)
{
    // A live portal keeps the target its children mounted into: re-registering the
    // id only points FUTURE portals elsewhere, and patching into a re-registered
    // element would diff this portal's slot range against another element's
    // children.
    return (recorded.Target, false);
}
```

[...]

PR #581, opened 2026-08-09
feat(velvet): let V.Portal take the container itself

### LINKED ISSUE #411: V.Portal has no overload taking a VisualElement / Ref<VisualElement> target directly, and FiberPortalRegistry's...
## What React does

`createPortal(children, domNode, key?)` takes the container node itself. There is no name, no registry and no namespace, so two independently mounted React roots in the same page cannot collide on a portal target, and any node you have a reference to (including one obtained from a ref) is a valid container.

## What Velvet does

`V.Portal(string targetId, ...)` is the only registry-target form; the element must first be published via `FiberPortalRegistry.Register(id, target)`, and `FiberPortalRegistry` is a `static class` holding a single `private static readonly Dictionary<string, VisualElement> _targets` for the whole process — not scoped to a `ReconcilerContext` or a mounted tree. A second `Register` with the same id logs a warning and overwrites. There is no overload accepting a `VisualElement` or a `Ref<VisualElement>`.

## What a developer hits

Two mounted trees (a very ordinary Unity setup: a HUD `UIDocument` and a menu `UIDocument`, or several Preview-window stories) that each register `"modal-root"` silently share one namespace — the second registration wins and the first tree's later portals mount into the other panel's element. A component library that ships a portal target id collides with the consuming app's. And the natural Velvet way to obtain an element — `refCallback` into a `Ref<VisualElement>` — cannot be used as a portal target at all; you must invent a globally unique string and register it out of band. Velvet's own code notes the consequence: the same-panel event bridge cannot be attached from `Register` because "that registry is a bare global static with no ReconcilerContext to guard duplicate attaches with".

## Where

/Users/apple/dev/velvet/Packages/com.velvet.core/Runtime/Reconciler/FiberPortalRegistry.cs, /Users/apple/dev/velvet/Packages/com.velvet.core/Runtime/Component/V.cs (Portal overloads ~line 1498), /Users/apple/dev/velvet/Packages/com.velvet.core/Documentation~/portals.md

Confirmed by reading the source. (1) /Users/apple/dev/velvet/Packages/com.velvet.core/Runtime/Reconciler/FiberPortalRegistry.cs is `public static class` holding `private static readonly Dictionary<string, VisualElement> _targets = new();` — one map for the whole process, not scoped to a ReconcilerContext, MountedTree or panel; its only reset is an `#if UNITY_EDITOR` RuntimeInitializeOnLoadMethod for domain reloads. `Register` warns `Id "{id}" is already registered. Overwriting.` and clobbers. (2) `grep "static PortalNode Portal"` over the package returns exactly two overloads in Runtime/Component/V.cs: `Portal(string targetId, …)` (line 1498) and `Portal(UILayer layer, …)` (line 1524). The latter targets a framework-managed layer *panel*, not an element the caller holds. There is no VisualElement or Ref<VisualElement> overload anywhere, and `PortalNode` (Runtime/Component/VNodeTypes.cs:262) carries only `TargetId` (string) and `Layer` (UILayer?) — "exactly one of the two is non-null" — so no element-valued target can even flow through the node. (3) The reconciler resolves by string on both paths: FiberNodeFactory.cs:450 `FiberPortalRegistry.Get(portalNode.TargetId!)` and FiberNodePatcher.cs ~828 `FiberPortalRegistry.Get(describe)`. (4) The global-ness is acknowledged rather than justified: ChildReconciler.cs:441 says the bridge is "Never attached from FiberPortalRegistry.Register itself: that registry is a bare global static with no ReconcilerContext to guard duplicate attaches with or later dispose the bridge through." (5) No documented deviation-with-reason exists: Documentation~/portals.md mentions the registry only in its title and the one-line sample `V.Portal("modal-root", children: …)`; grepping it for Register/registry/namespace/collide yields nothing else, and Documentation~/README.md only names "registry targets". (6) Every portal test fixture (SamePanelPortalBubblingTests.cs, PortalTests.cs, PortalContextInheritanceTests.cs, TimeSlicedReconcilerTests.cs, ReconcilerAbortSafetyTests.cs, ComponentStateReparityTests.cs) hand-picks a globally unique id string, working around the shared namespace. Two further divergences with no React counterpart, since React resolves no name at all: an unregistered id at mount logs "Target \"{id}\" is not registered. Children will not be rendered." and mounts a hidden placeholder that heals on the first patch after registration; and re-registering an id under a live portal does NOT move that portal's children — ResolvePortalTarget keeps `recorded.Target` and only future portals see the new element, so the id→element binding is late-bound at mount then frozen.

## What exists today

Only two portal forms exist. `V.Portal(string targetId, VNode?[] children, string? key)` resolves through the process-global `FiberPortalRegistry` (string → VisualElement), which the caller must populate with `FiberPortalRegistry.Register(id, element)` — so an element you hold can be a target, but only by publishing it under a name in one namespace shared by every mounted tree in the process. `V.Portal(UILayer layer, …)` and `V.WorldSpace(position, …)` avoid the registry entirely but target framework-created host panels (one per layer per reconciler, or a GameObject+world-space panel), not a container the caller supplies. There is no overload, prop, or PortalNode field accepting a `VisualElement` or `Ref<VisualElement>`.

## What already exists

FiberPortalRegistry.Register(string, VisualElement) / Get / Unregister / IsRegistered; V.Portal(string targetId, VNode[], string key); V.Portal(UILayer, VNode[], string key, PanelFocusOrder); V.WorldSpace(...); V.DragOverlay(...); Ref<T>.SetElement and BaseElementNode.RefCallback (how you get hold of a VisualElement in the first place)

Checked PublicAPI.txt (lines 267-270, 463-464, 741-744), Runtime/Reconciler/FiberPortalRegistry.cs, Runtime/NameKeyedRegistry.cs, Runtime/Component/V.cs (V.Portal/V.WorldSpace/V.DragOverlay), Runtime/Reconciler/ChildReconciler.cs (ResolveRegistryPortalTarget / ResolveLayerPortalTarget / ResolveWorldSpacePortalTarget), Runtime/Reconciler/FiberNodeFactory.cs CreateForPortalNode, Documentation~/portals.md, and the fixtures PortalTests / PortalContextInheritanceTests / SamePanelPortalBubblingTests / PortalLayersTests / FocusChainedPortalTests.

WHAT THE CLAIM GETS RIGHT (genuinely missing):
1. No V.Portal overload takes an element reference. The only two signatures are V.Portal(string targetId, ...) and V.Portal(UILayer, ...); PortalNode carries TargetId (string) and Layer (UILayer?) as an enforced one-of pair (FiberNodeFactory warns and mounts an inert placeholder if both or neither are set). There is no createPortal(children, container) equivalent — you cannot pass the VisualElement, nor a Ref<VisualElement> obtained from refCallback/Ref<T>.SetElement, to a portal.
2. The registry namespace is genuinely process-global and unscoped: FiberPortalRegistry is a `public static class` over a single `private static readonly Dictionary<string, VisualElement> _targets`. There is no per-reconciler, per-MountedTree or per-panel scoping; a second Register of the same id logs a warning and overwrites (NameKeyedRegistry.Set last-writer-wins), and the dictionary is only cleared by a RuntimeInitializeOnLoadMethod under UNITY_EDITOR (plus an internal test-only Clear()). ChildReconciler.cs:441 states this outright in a comment: "that registry is a bare global static with no ReconcilerContext to guard duplicate attaches with or later dispose the bridge through."

WHAT THE CLAIM GETS WRONG:
1. "A portal cannot target an element you hold" overstates it. You CAN portal into an arbitrary element you hold — including one captured by refCallback — via FiberPortalRegistry.Register("my-id", element) followed by V.Portal("my-id"). That is exactly what the fixtures do (SamePanelPortalBubblingTests creates a plain VisualElement, adds it to the window root, registers it as "evt-target"; PortalContextInheritanceTests registers "ctx-portal-target"). The capability exists; what is missing is the direct-reference ergonomics — the id round-trip through a global map.
2. "targets live in one process-global string namespace" is true only of the registry form, which is one of three portal forms. V.Portal(UILayer, ...) resolves a framework-managed layer host lazily at drain time, one host panel per layer per reconciler (ResolveLayerPortalTarget), and V.WorldSpace creates and owns its own GameObject + world-space panel host (ResolveWorldSpacePortalTarget). Neither touches FiberPortalRegistry or any string key, so the two common "render above everything" cases are already free of the global namespace.

s4k10503:
Closes #411.

`V.Portal` could only address a target an id had been published for. React's `createPortal` takes
the container itself, and every element a `refCallback` hands back is a valid container without
ever being registered — so the registry was a mandatory step for a case that does not need it, and
two subtrees choosing the same string collided across the whole process.

```csharp
V.Portal(targetId: "modal-root", children: …)   // registry, unchanged
V.Portal(layer: PortalLayer.Overlay, …)         // framework layer, unchanged
V.Portal(container, children: …)                // new — the element itself
```

`PortalNode.TargetId` and `Layer` were a one-of pair and are now a one-of triple with
`TargetElement`. A hand-built node setting none or more than one still warns and mounts an inert
placeholder.

## The container is not frozen at mount

A later render naming a different container moves the children — the old is emptied, the new
receives them — which is what `createPortal` does. `ReconcileKeying.CanPatch` gained a
`ReferenceEquals` term, so the move arrives as an unmount and a remount rather than a patch, and
child state, refs and effects do not survive it.

This differs from the registry form, whose id resolves at mount and is then held. That asymmetry is
pre-existing and out of scope here; `portals.md` states it as a deliberate difference. #412 is the
issue that revisits it.

## StrictMode

`FiberStrictMode` compares a per-render signature to catch a render body whose output depends on
mutated state. The portal term was the target id or the layer, so a body evaluating to a **fresh
container each render** produced an identical signature both passes — and under the semantics above
that body now remounts the entire portal subtree on every render, which is precisely the impurity
the gate exists to surface. The signature carries the container's identity.

## Verification

- `PortalElementTargetTests` — mount into the caller's container, the move on change, the patch on
  no change, and two containers not sharing a namespace.
- The patch-in-place case uses a `V.Div`. A `V.Text` mounts a pooled `Label`, and a remount returns
  it to a LIFO pool the very next rent pops, so a reference reading could not tell a patch from a
  remount there; with a `Label` the case passed on the broken behaviour too.
- `StrictModeDoubleInvokeTests` gained the fresh-container case, red without the signature term
  (`Expected log did not appear`) and green with it.
- EditMode 3781/3781, `assert_no_inconclusive.py` clean.

`Documentation~/portals.md` documents the form, the move semantics and the registry contrast; its
`README.md` index row is unchanged in shape.


s4k10503:
## Review round: 7 findings, 6 fixed, 1 rejected with evidence

Each finding was reproduced before it was touched. Full EditMode **3794 passed / 0 failed / 0 inconclusive** and PlayMode **161 / 0 / 0**, `assert_no_inconclusive.py` clean on both.

### HIGH — the bridge table pinned every dead container

Element portals route through the same-panel drain branch, which attaches the synthetic-bubbling bridge once per resolved target. Release happened only at `Reconciler.Dispose`, on a rationale written for a registered id — an element the app owns and that outlives any one portal. A container the tree itself renders does not: a keyed list whose rows each portal into their own `refCallback` container left one dead element and its ten callbacks pinned per unmount.

`FiberElementCleaner.CleanupPortal` now releases the bridge once no live `PortalState` entry resolves to that element. Counted by scanning `PortalState` rather than by a per-target counter, whose increment would have to sit outside the attach-once guard that skips every portal after the first.

RED without the fix, reading the bridge table across a mount/unmount cycle:

```
Given_a_mounted_portal_When_it_unmounts_Then_the_container_leaves_the_bridge_table
  Expected: (1, 0)
  But was:  (1, 1)
```

The mounted count is folded into the tuple so a bridge that never attached cannot satisfy the release reading on its own.

### MEDIUM-HIGH — the wrong comment on the patch-time resolver

`ResolvePortalTarget`'s header claimed `TargetElement` was not among the targets it resolves, sixteen lines above the branch that resolves it. Rewritten to enumerate all four ways a portal addresses a target, with the `CanPatch` fact stated once there instead of twice.

### MEDIUM — the DevTools dump printed an empty target

RED:

```
Given_APortalIntoAHeldContainer_When_TheTreeIsDumped_Then_TheContainerIsNamed
  Expected: String containing "held-container"
  But was:  "[Portal] target=
  <VisualElement #in-portal>
"
```

The line now carries the element's type and, where it has one, its name. All three arms of the one-of triple are pinned by a test.

### MEDIUM — the pasted StrictMode test

The `[Component]` attribute sat on the NUnit `[Test]` method, leaving `ImpureRender` and `FreshPortalContainerRender` unmarked. Both render bodies carry it again and the test moved into the render-double-invoke region.

### MEDIUM — "all three forms" / "four ways"

Corrected to four.

### LOW — the two-containers test was test 1 twice

Replaced with the two cases that were missing: two element portals sharing one container each keeping their own slot range, and an element-portal unmount taking only its own range out of a container that also holds a caller-owned child.

### LOW — `Assert.That(mounted, Is.Not.Null)`

Replaced with a reading that can fail: the two diagnostic passes build two containers and exactly one of them keeps the portal's children, so the committed pass's mount is the only one left behind.

---

### Rejected: enforcing `target != placeholder.parent` at the mount

The suggestion was to enforce this at the mount rather than leave it to a `Debug.Assert`. Implemented, it broke a supported feature. Nesting a portal inside another portal on the same target puts the inner placeholder inside the target by construction, so the predicate refuses it:

```
PortalTests.Given_OuterPortalContainsInnerPortalSameTarget_When_Mounted_Then_BothContentsCoexistInTarget
  System.ArgumentOutOfRangeException : Index was out of range.

PortalTests.Given_NestedPortalsSameTarget_When_OuterUnmounts_Then_WholeChainIsRemoved
  Inconclusive — Precondition: both outer and inner contents are mounted
  Expected: 3  But was: 2
```

The enforcement is reverted and the `Debug.Assert` restored as it was.

The finding underneath it is real, and broader than the predicate names. A portal's `SlotStart` is recorded after whatever the target already held, and only *another portal's* patch shifts it (`PortalSlotTracker.ShiftSlotStartsAfter`). An ordinary render of the target's own children does not. Measured, on a container that is an ordinary sibling rather than the portal's parent:

```csharp
// box mounted with one child "a"; a portal into box holds "p"
// then box's own children grow to [a, b] while the portal is mounted
Expected: "a|b|p"
But was:  "a|p|p"
```

The container's own new child is written over and the portal's child is duplicated — the same corruption the parent case was named for, from the same cause, with no warning emitted. So `target != placeholder.parent` is a proxy that both over- and under-approximates the real condition.

Fixing it properly means shifting the portal ranges on a target after an ordinary reconcile of that target, which touches the general child-reconcile path, needs a target-keyed index to stay off the hot path, and has to not double-count against `PatchPortalChildren`'s own shift. That belongs in its own change. For this PR the guide states the restriction: keep the container's own children out of Velvet's hands while a portal is mounted, and a nested portal on the same target is explicitly not that case.

### Also found, not fixed

`FiberNodePatcher`'s restored `Debug.Assert(target != placeholder.parent, …)` is a false positive for exactly the nested same-target portal above — no current test patches an inner nested portal, which is why it has never fired. It is pre-existing and left alone here.

s4k10503:
Review round 2. No HIGH survived — the bridge-release fix holds. Three claims were false and are corrected.

- **MEDIUM** — `portals.md` said "Nothing warns." of the container-with-its-own-children arrangement. An assertion does warn, in the editor: `Debug.Assert` is `[Conditional("UNITY_ASSERTIONS")]`, so the editor logs on every patch from the second render and a **player logs nothing and takes the corruption silently**. That is a sharper statement than the one it replaces and it is the one a reader needs.
- **LOW-MEDIUM** — `Reconciler.cs`'s dispose sweep still said same-panel targets are "ordinary, user-owned elements", the premise the fix layer itself overturned when it rewrote `SamePanelPortalBridges` to say a target is routinely one Velvet rendered. The rename sweep updated six comments and missed the one sitting on the code that consumes the table.
- **LOW** — the fixture summary claimed a per-form distinction the code does not make: `ReleaseBridgeIfLastPortalOn` keys on the target alone and never branches on form.

EditMode 3794/3794.

s4k10503:
Review round 3. The round-2 prose fix was itself wrong, so the sentence is deleted rather than corrected a third time.

- **HIGH** — round 2 replaced "Nothing warns." with "an assertion catches the arrangement in the editor". **No assertion is on that path.** `Runtime/` has six `Debug.Assert` sites and none covers a container gaining an own child ahead of a portal's range; the one that does fire is the mirror case, an own child *removed*. So round 1's "Nothing warns" was right, round 2 made it worse — a reader would take a silent Editor session as evidence the arrangement is safe — and CLAUDE.md says a sentence corrected twice and still wrong gets deleted. Deleted. The behaviour above it is what a reader acts on; which diagnostics happen to fire is an engine fact no test here holds.
- **MEDIUM** — `Reconciler.cs`'s dispose comment claimed the loop covers "one Velvet rendered above the subtree being torn down". It cannot: `ReleaseHostsAndScopes` runs only for the context-owning root, where nothing is above the tree. It also re-explained the rationale it names an owner for. Cut to the decision, with the owner named.
- **LOW** — the fixture summary asserted release behaviour for registry ids, which is true but which no case in that fixture exercises. Removed; an unpinned mirror is what the deleted sentence was.

EditMode 3788/3788.

PR #582, opened 2026-08-09
fix(velvet): let a second Mutate call run without cancelling the first

### LINKED ISSUE #420: A second `Mutate` call cancels the first and silently drops its callbacks
## What React does

TanStack Query v5 does not cancel mutations — `mutationFn` is not even handed an `AbortSignal` (only `queryFn` gets one). Calling `mutate` twice starts two independent `Mutation` instances; both run to completion and both fire the hook's `onSuccess` / `onError` / `onSettled`. The observer's `data` / `status` reflect the most recent one, and `useMutationState` / `useIsMutating` let you observe the set.

## What Velvet does

`RunMutationAsync` opens with `slot.Cts?.Cancel(); slot.Cts?.Dispose();` before installing a fresh CTS — "the latest call wins". Every later branch guards on `slot.Cts != cts` and returns without touching the slot or invoking `OnSuccess` / `OnError`. One `MutationResult` holds a single `Status` / `Data` / `Error` / `Variables`, so concurrent invocations are not individually observable.

## What a developer hits

A double-tapped Buy or Send button aborts the first request's token mid-flight while the server may already have committed it, and the first call's `OnSuccess` never runs — so a developer who puts "add the item to the inventory store" in `OnSuccess` loses the first purchase's write while the charge stands. Firing the same mutation for several rows of a list (save all loadouts) works in TanStack and silently keeps only the last one here. No constraint in Unity forces this; the cancellation token itself is a good Unity addition (UnityWebRequest abort), but auto-cancelling on re-entry is a semantic choice that is neither React's nor TanStack's and is not called out in the migration guide.

## Where

/Users/apple/dev/velvet/Packages/com.velvet.core/Runtime/Hooks/Hooks.cs (RunMutationAsync, ~L1752-1800); /Users/apple/dev/velvet/Packages/com.velvet.core/Runtime/Hooks/HookMutation.cs (MutationResult)

Confirmed by reading the source. `RunMutationAsync` (/Users/apple/dev/velvet/Packages/com.velvet.core/Runtime/Hooks/Hooks.cs, L1753-1806) opens with `slot.Cts?.Cancel(); slot.Cts?.Dispose();` before installing a fresh CTS (L1761-1764), so a second `Mutate`/`MutateAsync` cancels the first. Every later branch guards on `slot.Cts != cts`: on success L1776 returns the data without writing the slot and without invoking `OnSuccess`; on failure L1788-1794 returns/rethrows without invoking `OnError`; and L1782 `catch (OperationCanceledException) when (cts.IsCancellationRequested) { return default!; }` swallows the cancellation entirely, so a `MutationFn` that honours its token leaves even the `MutateAsync` caller with `default` and no exception. `MutationResult<TVariables, TData>` (Runtime/Hooks/HookMutation.cs) carries a single `Status`/`Data`/`Error`/`Variables`, and there is no `useIsMutating` / `useMutationState` equivalent in the package (grep for `IsMutating`, `MutationState` across Runtime/ and Documentation~ returns nothing).

Divergence from TanStack Query v5, precisely: (1) TanStack never cancels — `mutationFn` is handed no `AbortSignal`; Velvet cancels the superseded call's token. (2) In TanStack the superseded `Mutation` still runs to completion and still fires the hook-level `onSuccess`/`onError`/`onSettled` (the observer only detaches, so the per-call `mutate(vars, {onSuccess})` callbacks are what get dropped); Velvet drops the hook-level callbacks too. (3) TanStack lets you observe the whole set via `useMutationState`/`useIsMutating`; Velvet has no such hook, and its single result object makes concurrent invocations individually unobservable.

Not a documented deviation. `Documentation~/react-migration.md` §1-2a ("Async mutations — Hooks.UseMutation", L92-105) claims "TanStack Query v5 parity" but only for the throwing-`onSuccess`/`onError` semantics; it never mentions cancellation or supersession — "cancel"/"supersede" do not occur in the guide. The behaviour is only restated, never justified, in the comment at Hooks.cs L1761 ("Cancel any superseded in-flight mutation. The latest call wins.") and in the fixture summary at Runtime/Hooks/Tests/Editor/UseMutationHookTests.cs L26. The pinning test `Given_ConcurrentMutations_When_LatestCompletes_Then_FinalStateComesFromLatestCall` (L259-281) asserts only the winner's `Data`/`Variables`/`Status`; nothing asserts anything about the loser's callbacks, and the test has to call `first.TrySetCanceled()` to "drain the orphaned first call".

One nuance the claim understates: for the superseded `MutateAsync` caller, if the `MutationFn` ignores the cancellation token and completes, L1776 does return the data to that awaiter — so the caller's own await is not always starved. What is unconditionally dropped is the hook's `OnSuccess`/`OnError` and any trace in the shared `MutationResult`.

## What exists today

`Hooks.UseMutation` exists (Runtime/Hooks/Hooks.cs L1676, plus void-return and no-input overloads at L1717/L1736) and returns a render-stable `MutationResult<TVariables, TData>` with `Mutate` / `MutateAsync` / `Reset` and `Status`/`Data`/`Error`/`Variables`, matching TanStack Query's shape for the single-in-flight case — including the documented v5-parity behaviour for throwing `OnSuccess`/`OnError` handlers. What does not exist is TanStack's independent-`Mutation`-per-call model: there is one slot-level CTS and one result object, so concurrency is resolved as latest-call-wins instead of "both run, both notify". There is also no `useIsMutating` / `useMutationState` counterpart under any name. `MutationResult.Reset()` is documented as explicitly *not* cancelling in-flight mutations (HookMutation.cs), which is the only cancellation-adjacent behaviour the public docs describe.

s4k10503:
Closes #420.

`UseMutation.Mutate` cancelled the in-flight call before starting the next one. So a double-tapped
Buy button aborted a request the server may already have committed, and the `OnSuccess` that would
have recorded the order never ran — the first call's callbacks were dropped along with its token.

TanStack Query, which `UseMutation` is modelled on, does none of that. `mutationFn` is handed no
signal at all; every call runs to completion and delivers its own `OnSuccess` / `OnError`; the
observer's `Status`, `Data` and `Variables` show the most recent call. This PR adopts that.

## What changed

The slot holds a list of live `CancellationTokenSource`s rather than one, and a monotonic
`Generation`. A call captures its generation at start:

- observable slot writes (`Status`, `Data`, `Error`, `Variables`) happen only while
  `mine == slot.Generation`, so a slow first call cannot overwrite a fast second call's result;
- callbacks fire unconditionally, because the caller asked for that call's outcome and a later call
  says nothing about it.

Cancellation on unmount is unchanged and is why the tokens still exist: there, a request genuinely
does want cancelling, and every live one is cancelled rather than only the newest.

Each source is removed from the list and disposed in a `finally`, so a long-running mutation loop
does not accumulate them.

## Verification

Four cases, each red without the fix: both `mutationFn`s run to completion; the first call's
`OnSuccess` fires after the second has started; a slow first call resolving last leaves `Data` at
the second call's value; and unmount still cancels every live call.

EditMode 3778/3778, `assert_no_inconclusive.py` clean.

`Documentation~/react-migration.md` described the cancellation as a deliberate deviation and now
records the parity.

### DIFF
--- a/Packages/com.velvet.core/CHANGELOG.md
+++ b/Packages/com.velvet.core/CHANGELOG.md
@@ -13,7 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   its own `OnSuccess` / `OnError`, and `Status` / `Data` / `Variables` show the most recent — which is
   what TanStack Query does, and it hands `mutationFn` no signal at all. A double-tapped Buy used to
   abort a request the server may already have committed and then skip the `OnSuccess` that records
-  it. The token survives for unmount, where a request does want cancelling.
+  it. The token survives for unmount, where a request does want cancelling. A starting call also
+  clears `Data` along with `Error`, so a result belonging to the previous call is not read as this
+  one's while it is still pending.
 
 - A `TabIndex` or `DelegatesFocus` prop that a later render stopped declaring was written back as `0`
   or `false` rather than as the value the element was constructed with — neither of which is the
--- a/Packages/com.velvet.core/CHANGELOG.md
+++ b/Packages/com.velvet.core/CHANGELOG.md
@@ -10,8 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - A second `Mutate` call no longer cancels the first or drops its callbacks. Both run, each delivers
-  its own `OnSuccess` / `OnError`, and `Status` / `Data` / `Variables` show the most recent — which is
-  what TanStack Query does, and it hands `mutationFn` no signal at all. A double-tapped Buy used to
+  its own `OnSuccess` / `OnError`, and `Status` / `Data` / `Variables` show the most recent. Not
+  cancelling and showing the newest is what TanStack Query does, and it hands `mutationFn` no signal
+  at all; delivering the superseded call's callbacks is a deliberate deviation, since TanStack
+  detaches the observer and runs neither its handlers nor the hook's. A double-tapped Buy used to
   abort a request the server may already have committed and then skip the `OnSuccess` that records
   it. The token survives for unmount, where a request does want cancelling. A starting call also
   clears `Data` along with `Error`, so a result belonging to the previous call is not read as this
--- a/Packages/com.velvet.core/CHANGELOG.md
+++ b/Packages/com.velvet.core/CHANGELOG.md
@@ -10,10 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - A second `Mutate` call no longer cancels the first or drops its callbacks. Both run, each delivers
-  its own `OnSuccess` / `OnError`, and `Status` / `Data` / `Variables` show the most recent. Not
-  cancelling and showing the newest is what TanStack Query does, and it hands `mutationFn` no signal
-  at all; delivering the superseded call's callbacks is a deliberate deviation, since TanStack
-  detaches the observer and runs neither its handlers nor the hook's. A double-tapped Buy used to
+  its own `OnSuccess` / `OnError`, and `Status` / `Data` / `Variables` show the most recent — which
+  is what TanStack Query does, and it hands `mutationFn` no signal at all. A double-tapped Buy used to
   abort a request the server may already have committed and then skip the `OnSuccess` that records
   it. The token survives for unmount, where a request does want cancelling. A starting call also
   clears `Data` along with `Error`, so a result belonging to the previous call is not read as this
--- a/Packages/com.velvet.core/CHANGELOG.md
+++ b/Packages/com.velvet.core/CHANGELOG.md
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- A second `Mutate` call no longer cancels the first or drops its callbacks. Both run, each delivers
+  its own `OnSuccess` / `OnError`, and `Status` / `Data` / `Variables` show the most recent — which is
+  what TanStack Query does, and it hands `mutationFn` no signal at all. A double-tapped Buy used to
+  abort a request the server may already have committed and then skip the `OnSuccess` that records
+  it. The token survives for unmount, where a request does want cancelling.
+
 - A `TabIndex` or `DelegatesFocus` prop that a later render stopped declaring was written back as `0`
   or `false` rather than as the value the element was constructed with — neither of which is the
   right answer for every type. A `Label` is built out of the tab ring at -1, which shows once a
--- a/Packages/com.velvet.core/Documentation~/react-migration.md
+++ b/Packages/com.velvet.core/Documentation~/react-migration.md
@@ -99,13 +99,13 @@ TanStack Query's `useMutation` equivalent. Returns a handle with `Mutate` (fire-
 | `mutate(variables)` | `mutation.Mutate(variables)` |
 | `mutateAsync(variables)` | `await mutation.MutateAsync(variables)` |
 
-**Concurrent calls** (TanStack Query v5 parity): calling `Mutate` twice starts two runs, neither
-cancels the other, each delivers its own `OnSuccess` / `OnError`, and `Status` / `Data` / `Error` /
-`Variables` are one snapshot following the newest — so a double-tapped button does not lose the
-first call's follow-up write. `Status` / `Data` / `Error` / `Variables` are one
-snapshot describing the most recent call, and starting a call resets all four to that call's own —
-`Data` included, so a pending call never shows the previous one's result. The `CancellationToken` handed to `MutationFn` is still cancelled when the
-component unmounts, which is what a Unity web request wants.
+**Concurrent calls** (TanStack Query v5 parity on cancellation and on the snapshot): calling
+`Mutate` twice starts two runs, neither cancels the other, each delivers its own `OnSuccess` /
+`OnError`, and `Status` / `Data` / `Error` / `Variables` are one snapshot following the newest — so
+a double-tapped button does not lose the first call's follow-up write. Starting a call resets all
+four to that call's own — `Data` included, so a pending call never shows the previous one's result.
+The `CancellationToken` handed to `MutationFn` is still cancelled when the component unmounts, which
+is what a Unity web request wants.
 
 **Callback error semantics** (TanStack Query v5 parity):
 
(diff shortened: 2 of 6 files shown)


s4k10503:
Review round 1, three findings, all fixed:

- **HIGH** — `Dispose()` walked `Live` while `Cancel()` resumed the mutation's continuation synchronously, whose `finally` removes from that same list. One in-flight mutation with a token-honouring `mutationFn` threw out of unmount and aborted the enclosing reconcile. Now snapshots and clears first. Pinned by `Given_AMutationInFlight_When_TheComponentUnmounts_Then_TheUnmountCompletes`.
- **HIGH** — `Given_ConcurrentMutations_When_LatestCompletes_Then_FinalStateComesFromLatestCall` passed on `main` too: its `mutationFn` returned a bare `UniTaskCompletionSource` task, which ignores the token, so the old cancel was a no-op and the term claiming to separate completion from cancellation separated nothing. Both concurrency fixtures now go through `AttachExternalCancellation(ct)`.
- **MEDIUM** — the parity claim was false for `Data`: a starting call reset `Status`, `Variables` and `Error` but left the previous call's `Data` in place, so a pending call read as having produced the older result. v5 resets it. Now cleared at start and pinned by `Given_ASucceededMutation_When_AnotherStarts_Then_DataIsNotTheOlderResult`; the guide and CHANGELOG say so.

EditMode 3787/3787.

s4k10503:
Review round 2 on the fix layer. Four fixed, two filed.

- **MEDIUM** — `MutationResult.Data`'s XML doc still said "result of the last **successful** mutation", which round 1's `Data = default!` falsified and DocFX publishes. It now describes the most recent call.
- **LOW, but a throw** — the `finally` disposed unconditionally while `Dispose()` clears `Live` before cancelling, so with A and B in flight where cancelling A settles B, unmount reached `Cancel()` on an already-disposed source and threw `ObjectDisposedException` out of the reconcile. `if (slot.Live.Remove(cts)) cts.Dispose();` — the removal's result is now what says who owns the disposal, and the comment on `Dispose()` says that is why it clears first.
- **LOW** — that same comment claimed the `finally` "removes the entry being walked"; the loop walks a snapshot and the `Remove` returns false. Round 1 wrote it and round 1's own change made it false.
- **LOW-MEDIUM ×2** — two gating `Assume`s were doing the discriminating work. Deleting `slot.Result.Data = data` from the success path left `Is.EqualTo((Pending, 0))` passing on an inconclusive, and the unmount case passed vacuously if the mutation never started. Both folded into their assertions as leading terms.

Filed rather than fixed, both pre-existing and outside this PR's change: `ResetMutation` does not bump `Generation`, so a `Reset()` during an in-flight call is overwritten when that call lands; and `Data` is written before `OnSuccess` runs, where TanStack dispatches success only after it returns, so a throwing handler leaves `Status == Error` with `Data` populated. Both are real and neither is something this PR introduced.

EditMode 3787/3787.

s4k10503:
Review round 3. The HIGH was mine and the claim was simply wrong.

- **HIGH** — three files called the whole concurrent-call behaviour TanStack v5 parity. It is not. v5's `MutationObserver.mutate` detaches the observer from the current mutation before building the new one, and `#notify` gates on `#mutateOptions`, which the second call overwrites — so **neither the superseded call's own callbacks nor the hook's run**. Three of the four halves are parity (no re-entry cancellation, no signal to `mutationFn`, newest-call snapshot with `Data` cleared at start); firing the superseded call's callbacks is a deliberate deviation, and it is the half this PR exists for. All three sites now say which is which.

  The deviation is the point, and saying so is more useful than the label was. The failure being fixed is a double-tapped Buy whose first `OnSuccess` never ran while the server had already committed. v5 can afford to drop those callbacks because a React app re-reads from a query cache; a caller here asked for that call's outcome, and a later call says nothing about whether it happened.

- **MEDIUM** — round 2's dispose fix (`if (slot.Live.Remove(cts)) cts.Dispose();`) had no test that goes red without it. It differs from round 2 only when `Remove` returns false, and disposing an already-disposed `CancellationTokenSource` returns quietly, so one in-flight call passed either way. Added a case with **two** calls in flight where cancelling the first settles the second, with the started-call count as the leading term so it cannot pass without exercising the ordering.

s4k10503:
Review round 4. **The round-3 relabel was backwards and is reverted.**

v5's `Mutation.execute` awaits `this.options.onSuccess` / `onSettled` / `onError` itself and consults no observer list, so a superseded mutation **does** fire the callbacks passed to `useMutation({ onSuccess, onError })`. What `removeObserver` plus the `#mutateOptions` overwrite suppresses is only the **per-call** form, `mutate(vars, { onSuccess })` — which Velvet has no equivalent of: `Mutate`/`MutateAsync` take variables only, and the handlers come from the `MutationOptions` given to `UseMutation`, which this guide's own table maps to `useMutation({ onSuccess, onError })`.

So the original claim was right and round 3 made it wrong. Left standing, a reader would have ported a suppression React Query does not have. Restored in all three files.

**There is a real divergence, and it is not the one I named.** `UseMutation` rewrites the slot's handlers every render and a call reads them at completion, so a call that started before a re-render runs the **newest** `OnSuccess`. v5 keeps the options snapshot the mutation was built with — `setOptions` forwards only while the mutation is still current and pending. Named in `react-migration.md` now.

The round-3 test survives review: it does reproduce the ordering (`Dispose` snapshots and clears `Live`, `cts1.Cancel()` settles call 2 synchronously, whose `finally` re-enters while the loop still holds `cts2`), it is RED without the `Remove` guard with the `ObjectDisposedException` reaching `mounted.Dispose()`, and `bothStarted` does what its comment claims.

EditMode 3788/3788.

s4k10503:
Review round 5. The divergence sentence is **deleted**, not corrected a fourth time.

Its history: round 3 called the whole concurrent-call behaviour a deviation (wrong — v5's `Mutation.execute` awaits `this.options.onSuccess` itself). Round 4 reverted that and named a different divergence — that Velvet runs the newest render's handler where v5 keeps the mutation's build-time options. **Also wrong.** `useMutation` runs `observer.setOptions(options)` every render, and `MutationObserver.setOptions` ends by forwarding to `this.#currentMutation.setOptions(this.options)` while that mutation is still current and pending; `Mutation.execute` late-reads `this.options.onSuccess`. So a pending v5 mutation takes the newest options on every render and freezes only at supersession — in the sequence the sentence named, v5 runs the newest handler too.

A divergence does exist, in exactly one ordering: the re-render landing **after** supersession (`Mutate(v1)`; `Mutate(v2)`; re-render; `v1` completes → v5 runs the handler captured at supersession, Velvet the newest). That is not the trigger the sentence named, and I am not writing a fourth version of it. CLAUDE.md: a sentence corrected twice and still wrong gets deleted. This one was wrong three times, in three different directions, and nothing in the code needed it.

What stays is what survived attack in both directions: `Mutation.execute` consults no observer list, so a superseded mutation's hook-level callbacks fire; only the per-call `mutate(vars, {onSuccess})` form is suppressed, and Velvet has no per-call form. The `Status` / `Data` / `Error` / `Variables` sentence is back in the parity paragraph it was stranded out of.

PR #583, opened 2026-08-09
tooling(velvet): refuse a CHANGELOG entry filed into a released version

s4k10503:
No issue: found by making the mistake. The portal branch's CHANGELOG entry went into the `### Added`
of **2.1.0** — a version released the day before — and nothing said so. The diff showed a bullet
appearing under `### Added`, which is what a correct entry looks like too; only the heading forty
lines above distinguished them, and it is not in the hunk.

Both halves of the result are wrong. `scripts/release/release_notes.py` rebuilds a note from a
CHANGELOG section on demand, so repairing the v2.1.0 release would have published a feature that
version does not contain — and the version that does contain it would ship without mentioning it.

Nothing existing catches this. `test_release_notes.py` asks whether each version can produce a
complete note; a closed section that gained a stray entry still has its Highlights block and still
produces one. The note is simply false, and no check reads truth.

## The check

A `PreToolUse` refusal on `Edit`/`Write` of a `CHANGELOG.md`, at the moment of insertion — by the
time a branch is pushed the entry has been written, reviewed, and cited as evidence.

The reading is a dated section's non-blank line count, before and against the proposed text. Only
growth is refused, so the three ways this mistake gets undone all still run: rewording a closed
entry, moving one out of a closed section, deleting one. A blank line is not an entry.

An `old_string` that does not occur is left alone — the edit fails on its own, and a guard that
guesses at where a failed edit would have landed is guessing.

## Verification

Ten cases, driven against the repository's own CHANGELOG: the mistake itself refused for `Edit` and
for `Write`; an addition under `## [Unreleased]` allowed for both; a reword, a whole-bullet
deletion, and a blank line inside a closed section allowed; a stale `old_string` allowed; a
`CHANGELOG.md` outside this repository allowed.

The first draft used `^` without `re.MULTILINE`, so it found no headings at all and allowed
everything — which is the failure mode a guard has: silence is indistinguishable from a pass. It
was caught by the case that exercises the mistake, which is why that case exists rather than only
the negatives.


s4k10503:
Review round 1. Two HIGH, and the guard was live in the checkout, so both were armed.

- **HIGH** — a version being CLOSED (`## [Unreleased]` → `## [X.Y.Z] - DATE`) newly dates a whole section, and `before.get(v, 0)` read that as growth from zero. Every release was refused. The second HIGH is the same one from the other side: the refusal advised opening an `## [Unreleased]` section, and doing so still exited 2 — no in-band way to clear it. Only versions dated on **both** sides are compared now, so closing a version is not an addition to it.
- **MEDIUM-HIGH** — a line count is not a proxy for an entry being filed. One edit rewrapping a released entry 4→3 lines while appending a bullet netted zero and was allowed. The reading is the **bullet** count now, which a rewrap does not move.
- **MEDIUM** — the same reading made the docstring's reword claim false (a reword that wraps to one more line was refused). True as written now, and pinned.
- **MEDIUM** — basename-only scoping refused a `CHANGELOG.md` outside this repository. Scoped to the project root.
- **MEDIUM** — the heading pattern restated a grammar `release_notes.py` owns and a two-space `##  [2.1.0]` slipped a bullet past it. Loosened to the widest form that still finds every heading that parser accepts, and the comment names the owner instead of restating it.
- **LOW** — `closed_spans()` and `owner()` were dead after the rewrite; deleted.

Eleven cases now, driven against this repository's own CHANGELOG: the motivating mistake refused for `Edit` and `Write`; a rewrap paying for an addition refused; closing a version allowed three ways including the release skill's exact edit; an `## [Unreleased]` addition, a reword, a deletion, a foreign CHANGELOG and a stale `old_string` all allowed.

The remaining LOW stands: a `sed -i` or heredoc rewrite bypasses an `Edit`/`Write` matcher entirely. It mattered more when the HIGH made bypassing the only way to release; it is now the ordinary limit of a tool-input guard.

s4k10503:
Review round 2. The fix layer carried a HIGH of its own, and it is the third silent-pass on this file in three commits.

- **HIGH** — round 1's repository scoping tested containment under the project root, which turned the guard **off in every `git worktree`** — where this repository does essentially all of its branch work. The motivating edit replayed in `wt-portal` returned 0 while the identical edit in the main checkout returned 2. Scoping is by the git common dir now, which every worktree of a repository shares and a foreign checkout does not. Both are driven as cases.
- **MEDIUM** — the both-sides-dated rule left the heading itself unguarded: renaming `## [2.1.0] - date` to `## [2.1.1] - date`, or stripping the date, carried any payload under it out of the comparison. A released heading that stops being one is refused separately now.
- **MEDIUM-HIGH** — the PR body described the line-count mechanism the second commit deleted. Rewritten.
- **LOW, standing** — no committed test. `.claude/hooks/` has no harness and `test.yml` does not cover the path, which is exactly why three silent-pass bugs shipped here in three commits and every one was found by a reader rather than a run. Filed as its own issue; this guard should not merge ahead of it, and I would rather that be explicit than implied.

Seven cases driven: the mistake refused in the main checkout **and** in a worktree; a date-strip and a version-rename refused; closing a version, an `## [Unreleased]` addition, and a foreign repository's CHANGELOG all allowed.

s4k10503:
## Review findings addressed

All three were reproduced first, driving the hook at `ac8b648` with real payloads in throwaway git repos.

### (1) The guard allowed the exact edit it exists to stop — confirmed, fixed

Inserting a fabricated `## [2.1.0] - 2026-08-09` block (its own Highlights and Fixed entries) immediately above the real one:

```
--- A duplicate released heading fabricated above the real one: exit 0
--- An empty duplicate released heading above the real one: exit 0
```

`counts[version] = ...` inside the heading loop, last write wins, no growth, no `unmade`.

Summing would not have closed it. The second case above is the reason: `extract_version_section` breaks at the FIRST heading matching the version and ends at the next `## [`, so a **bullet-free** duplicate above the real one is the whole published note, and a sum sees no growth. A version heading written a second time is now refused on its own, before either section is read — and only when the edit is what introduces it, so a file that already carries a duplicate can still be repaired in band.

### (2) The refusal message asserted a mechanism that had not happened — confirmed, fixed

```
--- The whole released 2.0.0 section deleted: exit 2
    Refusing this CHANGELOG.md edit: it stops 2.0.0 being a released section, by renaming its version or removing its date.
```

Neither happened. The message now states what the proposed text carries and names an action; the inline comment lists all three routes; and the docstring bullet ("Only growth is refused, so deleting from a released section and moving an entry out of one both still run") now separates deleting an entry — still allowed — from deleting the section, which is not.

### (3) `common_git_dir` failed open — confirmed, fixed

```
--- growth edit, PATH holds no git: exit 0 in 30 ms
--- growth edit, git hangs past the timeout: exit 0 in 5032 ms
--- growth edit, CLAUDE_PROJECT_DIR names no repository: exit 0 in 43 ms
```

Now via `lib/repository.git`, and an unanswerable scoping question runs the check rather than standing down:

```
--- growth edit, PATH holds no git: exit 2 in 28 ms
--- growth edit, git hangs past the timeout: exit 2 in 5042 ms
--- growth edit, CLAUDE_PROJECT_DIR names no repository: exit 2 in 61 ms
```

### The 1-for-1 substitution

Also confirmed allowed at `ac8b648` (`exit 0`), and closed by borrowing `test_release_notes.py`'s item comparison: top-level list items split on a line starting a new bullet, each collapsed to one line. This refuses a reword of a released entry too — a substitution and a reword are the same text change, and no reading separates them — which is the accepted cost, stated in the docstring, with the refusal naming what a genuine reword should do instead. A rewrap still passes (`exit 0`), which is the control that says the comparison is not simply refusing everything.

### The mistyped-heading refusal

The `unmade` message now ends in an action rather than a bare statement. It does not distinguish a correction from an erasure, and I did not try to: the three previous attempts at a sentence about this were each wrong in a new direction.

## Controls

Against the tracked CHANGELOG, in a temp repo: closing `## [Unreleased]` into `## [2.2.0] - 2026-08-10` → exit 0; adding under `## [Unreleased]` → exit 0; and, on the post-release state with no Unreleased heading left, opening the first one above the newest release → exit 0. The tracked file carries no duplicate headings and reads 10 released versions.

## The required test

`HookWiringCoverageTests` now poses an Edit. Its three existing payloads all return at `if tool not in ("Edit", "Write")`, so nothing this branch added had ever been executed by a test. The new case builds a checkout, a worktree of it, and a growth edit inside the worktree — where scoping by containment and scoping by the shared git dir disagree.

RED against the hook as it stood at `c6284a8` (blob `febb370`):

```
Given_ACheckoutAndAWorktreeOfIt_When_TheClosedVersionGuardIsPosedAnEditInTheWorktree_Then_ItRefuses
  an entry filed into a released section is refused in the project directory and allowed everywhere the work happens:
  Expected: (True, True, 2)
  But was:  (True, True, 0)
```

The two `True`s are the arrangement — git linked the worktree, and it sits outside the checkout — folded into the same comparison, so the exit code cannot be read off a worktree that was never one.

## Suites

Full EditMode, headless, graphics on: **3785 passed, 0 failed, 0 inconclusive, 0 skipped**; `assert_no_inconclusive.py` clean. No compile errors in the log.

No `Documentation~` guide or markdown file references this hook or the rule it enforces, so nothing to update. No `[Unreleased]` entry: the package CHANGELOG records user-visible package behaviour, and this is contributor tooling.

s4k10503:
Both findings addressed at `968a782`.

## MEDIUM — the one new case was green on the parent

Confirmed by replay. The shipped payload (checkout + `git worktree add` + `- As shipped.` → `- As shipped.\n\n- Smuggled in.`, `CLAUDE_PROJECT_DIR` = the checkout) exits 2 on `3c79289` **and** on `3c79289^`, and `Linked`/`Outside` are properties of the arrangement, so the whole tuple held unchanged. It pins the worktree scoping `ac8b648` landed. Kept, because nothing else pins that.

Four cases now separate the two commits. Each was measured against the parent hook, and each also against a cut of the mechanism it names:

| case | parent | fix | mechanism cut |
|---|---|---|---|
| second `## [1.0.0]` heading carrying **no items** | 0 | 2 | 0 |
| one entry out, one in (bullet count unchanged) | 0 | 2 | 0 |
| project directory git cannot place, growth edit | 0 | 2 | 0 |
| refusal for a section deleted outright | message names a rename | does not | — |

**One correction to the suggested payload.** The `### Fixed` / `- Fabricated.` block does discriminate against the parent, but it is caught by *two* mechanisms in the fix: `released()` accumulates per version across both headings, so the fabricated bullet is an entry the section did not carry and the item reading refuses it on its own. With `repeated()` cut out it still exits 2 — it would have pinned the duplicate-heading fix in name only. A second heading carrying **no items** exits 0 with that branch cut, which is also the case the commit message calls the reason summing would not have closed it.

Parent-hook failure text, verbatim:

```
Given_AnEditWritingASecondHeadingForAReleasedVersion_... -> Failed
  Expected: 2
  But was:  0

Given_AnEditSubstitutingTheOnlyEntryOfAReleasedSection_... -> Failed
  Expected: 2
  But was:  0

Given_AProjectDirectoryGitCannotPlace_... -> Failed
  Expected: (False, 2)
  But was:  (False, 0)

Given_AnEditDeletingAReleasedSectionOutright_..._Then_ItDoesNotNameARename -> Failed
  Refusing this CHANGELOG.md edit: it stops 1.0.0 being a released section, by
  renaming its version or removing its date.
  Expected: (2, False)
  But was:  (2, True)
```

A fifth case pins the half that stands down on purpose, and is green on the parent for that reason.

## LOW — the claim was false, and the target half stays fail-open

Reproduced with `lib/repository.git` stubbed to answer for the project dir and not for the target — faithful to both hazards, since that reader returns `None` for a timeout and for a non-zero exit alike:

```
target really is a worktree of the project: True
project dir answers                      : True
target path answers                      : False
in_scope on a growth edit there          : False
```

So the growth edit is allowed. The new commit message says the claim is true of the project directory alone.

**The target half should not fail closed.** The two `None`s are not the same fact. For the project directory `None` has no ordinary reading — the session's project dir is a checkout of this repository, so an unanswerable one is the instrument failing. For the target it is *also* git's ordinary answer for a path in no repository, and `lib/repository.git` cannot separate that from a reading that failed. Failing closed there would leave nothing out of scope but a path git positively names another repository for, and every `CHANGELOG.md` outside one would be this guard's to refuse. The asymmetry is now in the docstring and pinned by the fifth case, which goes red if the target half is changed to `common_git_dir(here) in (mine, None)`.

The targeted fix that *would* close the timeout/`safe.directory` hole needs `lib/repository.git` to distinguish "ran and said no" from "could not run" — a contract change across every caller, and its own pinned cases. Not done here.

## Merge with `origin/main`

`main` landed the `HOOK_TOOLS` contract in the meantime, which this guard predated: it gated on a literal tuple and rode the `Edit|Write|NotebookEdit` matcher, routing a tool it does not act on. It now declares `{"Edit", "Write"}`, reads that set on the gate line, and takes a matcher of its own, with a `GatePayloads` row it decides about. The gate probes also pin `CLAUDE_PROJECT_DIR` to the fixture's scratch directory — scoping reads it, so otherwise the row's verdict would turn on whether the session running the suite exported one.

Incidentally, the parent hook fails that probe too: with the project dir pointed at a non-repository it stands down, which is the same fail-open defect from a second direction.

## Suites

EditMode 3822 passed / 0 failed / 0 inconclusive. PlayMode 161 / 0 / 0. `assert_no_inconclusive.py` clean on both. No `-nographics`.

No `Documentation~` guide or `CONTRIBUTING.md` section names this guard; `CONTRIBUTING.md` already documents the `HOOK_TOOLS` contract this change conforms to. No CHANGELOG entry — contributor tooling, not user-visible package behaviour.

s4k10503:
## Review round: the closed-version guard's reading

### What the reading has to be, before the patch

The guard's question is: **does the proposed text add anything to the body `release_notes.py` would publish for a version that is already released?** A count over a text range can express that, but only under four conditions — and each defect this file has shipped is exactly one of them broken:

1. **The range must be the publisher's range**, not a second parse of the same file. Two grammars have to agree exactly to be sound: a heading only one of them sees moves text across a version boundary for that one and not the other, and a disagreement in either direction can leave a released section's text uncompared and published anyway. Nothing holds two grammars to that.
2. **The units must partition the range** — every non-blank line in exactly one unit — rather than sample it. Whatever the units do not cover, the guard allows, and a diff never says which part that is.
3. **The ranges compared must be the ones published on both sides.** A range that was not published before is not one an edit can smuggle into.
4. **The files compared must be this repository's**, by the identity a worktree shares with its checkout.

Mapped onto the branch:

| commit | broken condition | effect |
|---|---|---|
| `7881d4d` | 3 | a line count compared against sections never published — closing `## [Unreleased]` into `## [X.Y.Z] - DATE` was refused. Measured: exit 2 on a legitimate release close |
| `7881d4d`…`f51771c` | 1 | the guard's heading pattern never matched `release_notes.py`'s. `^## \[` (7881d4d) misses a heading written with two spaces; `^##\s+\[` (c6284a8 onward) spans a newline and finds one the publisher does not |
| `c6284a8` | 4 | scoping by containment under the project directory — off in every worktree |
| `c6284a8` | 2 | `bullets()` assigned where it had to sum, so a version with two headings had one counted. Measured: a bullet smuggled under a duplicate heading exits 0 |
| `3c79289` | 2 | units are the column-0 list items only. Everything above a section's first one is unread |

So the answer to "can a line-or-item count over a text range express it": **a count cannot** — that was `c6284a8`'s other hole, a 1-for-1 substitution is invisible to a number — but a **multiset of covering units over the publisher's range** can, and that is the only shape that closes all of 1–4 at once. What it needs is not a wider regex; it is that the units be the publisher's own lines.

### The fix

`release_notes.VERSION_HEADING` delimits the range and `release_notes.unwrap_soft_breaks` unwraps it, both imported from the module that publishes — the same arrangement `merge_onto_unpublished_release.py` uses for `published_check`. Every non-blank line of the result is one unit, with the indent kept (it is what nests a line) and the wrap dropped (the join has already removed it). The guard now has no grammar of its own for a second one to disagree with, and no line falls between readings.

**HIGH** and **LOW** are both closed by that single change; **MEDIUM** is the `items()` docstring, which no longer exists.

### RED evidence

Repro on the repository's own `CHANGELOG.md`, an indented bullet above 2.1.0's first Highlights entry:

```
f51771c (parent)   EXIT 0
5ea1eca            EXIT 0
3c79289            EXIT 0   <- regression introduced here
c6284a8            EXIT 2
ac8b648            EXIT 2
```

Both new cases measured against the parent hook `f51771c` with the fixture the tests use:

```
nested above first bullet    parent f51771c  exit=0    fixed  exit=2
heading split across lines   parent f51771c  exit=0    fixed  exit=2
```

Fixed-side message, both cases: `Refusing this CHANGELOG.md edit: 1.0.0 comes out of it carrying a line it does not carry now, and a release date has closed that section.`

Cut matrix, to show each case dies at its own mechanism and not at the other's:

| case | intact | cut: loose heading grammar | cut: item-only units |
|---|---|---|---|
| entry nested above first column-0 bullet | refused | refused | **allowed** |
| entry below a heading split across lines | refused | **allowed** | refused |
| plain growth (control, already pinned) | refused | refused | refused |

### A claim in the review I could not reproduce

"`^` without `re.MULTILINE` (found no headings, allowed everything)" is not one of the defects this branch shipped. `7881d4d`, its first commit, already carries `re.MULTILINE`; `gh api repos/s4k10503/velvet/pulls/583/commits` lists nine commits and none precedes it. Four of the five named defects shipped here. The fifth slot is filled by the item the review filed as LOW: the heading grammar has disagreed with the publisher's since the first commit, in both directions in turn.

### Suites

Measured on the pushed tree, `origin/main` merged in, headless, no `-nographics`, machine otherwise idle:

- EditMode: total 3852, passed 3852, failed 0, inconclusive 0, skipped 0
- PlayMode: total 161, passed 161, failed 0, inconclusive 0, skipped 0 (Component 4, Hooks 7, Preview 1, Reconciler 82, StarterSample 3, Styling 64)
- `assert_no_inconclusive.py` on both: `checked 1 test run(s): no inconclusive cases`
- 0 `error CS` in either log

### Documentation and CHANGELOG

No `Documentation~` guide or `CONTRIBUTING.md` section describes this hook — the branch documents it in its own docstring, which this change rewrites. No CHANGELOG entry: a PreToolUse guard is contributor tooling, not user-visible package behaviour, and the branch adds none for the same reason.

### Found, not fixed

- **A released heading's date can be changed.** `## [1.0.0] - 2026-01-01` → `- 2026-01-02` is allowed. `extract_version_section` excludes the heading line, so nothing published changes — but the CHANGELOG's own claim about when a version shipped does, silently. Outside the reading by construction; closing it means comparing the heading text too.
- **A whole new dated section can be fabricated.** Inserting `## [1.5.0] - 2026-03-03` with entries under it is allowed, because only versions dated on both sides are compared — the same rule that keeps release-closing possible. Reaching an existing released version by that route is caught by the duplicate-heading reading.
- **Blank-line-only edits inside a released section are allowed.** They carry no content, so they are not units; they do change the note's spacing.
- **`import release_notes` at module load has the same exposure as the two imports beside it.** A missing module raises, the hook exits 1, and the edit proceeds — which `lib/repository`'s docstring calls a silently deleted guard. `import repository` here and `import published_check` in `merge_onto_unpublished_release.py` are identical in this respect, so I left all three consistent rather than hardening one.

Issue #574, opened 2026-08-09
HELD_TOOLS and the settings matcher are a hand-kept pair, and a trim to either is silent


s4k10503:
`edit_while_a_ready_pr_sits.py` decides which tools it holds from `HELD_TOOLS`, and `.claude/settings.json` decides which tools reach it from `"matcher": "Edit|Write|NotebookEdit"`. The two are a hand-kept pair and nothing compares them.

Drift scenario, and it is the one a maintainer is most likely to produce: trim `"NotebookEdit"` out of `HELD_TOOLS`. The matcher still routes `NotebookEdit` to the hook, the hook returns 0 for it, notebook edits proceed while a green pull request sits — and the whole suite stays green. The reverse drift is quieter still: drop it from the matcher and the set keeps claiming a tool the hook never sees.

This surfaced while correcting two sentences in the hook that said only `Edit` and `Write` are held, three lines above the set that also holds `NotebookEdit`. The prose is fixed; the two-place fact is not.

## What would close it

`HookWiringCoverageTests` already reads both files — it pairs every hook against `.claude/settings.json` by path, and reads each hook's source for `HOOK_SCOPE`. The missing case asserts that a hook declaring a tool set is registered on a matcher naming exactly those tools.

Measure before writing it. `edit_while_a_ready_pr_sits.py` is the only hook that filters by a declared set today — every other guard reads `tool_name` inline against a literal — so a guard derived from "hooks that declare a set" has one subject, and one subject is how a guard ends up asserting the shape of the thing it was written against rather than a rule. Worth checking whether the rule is better stated the other way round: every Bash-matcher hook checks `tool_name == "Bash"`, every Edit-matcher hook checks membership, and a mismatch between a matcher's alternation and what the hook actually admits is the general form.

PR #584, opened 2026-08-09
tooling(velvet): say when a deferral was ignored rather than reading as absent

s4k10503:
No issue: found by writing one. Three pull requests were held with a stamp of *now + 45 minutes*,
read as the moment the deferral should expire. `deferred()` bounds the age below as well as above,
so a future stamp is rejected — and a rejected deferral and an absent one are the same `None`. The
Stop guard fired again with the identical text, saying nothing about the three entries that had
just been written to answer it.

Every ingredient of that is in the guidance. `deferrals.py` documents the field as `$(date +%s)`;
the guard's own message prints a literal epoch of the current second. Reading it as an expiry is
not the guidance being wrong, it is the guidance being silent about which of two readings it takes
— and silence is what a malformed entry then returns.

## The change

`deferrals.unusable(key)` returns why the newest entry for a key cannot be honoured: a last field
that is not an epoch second, or a stamp in the future. Both Stop guards report it, alongside the
held list they already print.

Expiry is deliberately not reported. That one is the design working — it ends with the guard firing
and the reason being re-stated, which is the whole point of the TTL.

Both guards are wired because `deferrals.py` states the invariant that they read one file under one
expiry, so a format fact landing in one and not the other is the drift it exists to prevent.

## Verification

Five inputs against `unusable`: a future stamp and a non-numeric last field both reported; a live
entry, an expired one, and a key with no entry all silent. Then the real guard driven end to end
against a future-stamped deferral for an open PR — it now names the entry and why it was ignored,
where before the entry produced no output at all.

### DIFF
--- a/.claude/hooks/lib/deferrals.py
+++ b/.claude/hooks/lib/deferrals.py
@@ -64,9 +64,11 @@ def deferred(key, now=None):
 
 
 def unusable(key, now=None):
-    """Return why the newest entry for `key` cannot be honoured, or None — which covers both a live
-    entry and no entry at all. Ask `deferred` which of those it is; this one answers only "written,
-    and rejected".
+    """Return why the newest entry for `key` was rejected outright, or None.
+
+    None is not a state: an entry that is live, one that has expired, and no entry at all all take
+    it, and `deferred` separates only the first from the other two. Nothing distinguishes an expired
+    entry from an absent one, here or anywhere, for the reason below.
 
     A rejected deferral and an absent one both make `deferred` return None, so writing an unusable
     one reads as having written nothing — the guard fires again with the same text and the entry
--- a/.claude/hooks/lib/deferrals.py
+++ b/.claude/hooks/lib/deferrals.py
@@ -20,6 +20,16 @@ DEFERRALS = Path.home() / ".velvet-pr-deferrals"
 TTL = 2700
 
 
+def epoch(field):
+    """The field as an epoch second, or None. `str.isdigit` is true of characters `int` rejects, and
+    the caller that reads a malformed entry runs before the one that would have raised — an exception
+    out of a Stop guard exits 1, which the harness treats as non-blocking, so the guard turns off."""
+    try:
+        return int(field)
+    except ValueError:
+        return None
+
+
 def deferred(key, now=None):
     """Return (reason, minutes) when a live deferral covers the key, else None.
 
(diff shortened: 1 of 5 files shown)


s4k10503:
Review round 1. Six findings, five fixed; the sixth is filed rather than fixed.

- **HIGH** — `unsettled_pr.py`'s unreadable-`gh` path advised arming a `backlog` deferral, and `main()` only ever queried PR-number keys — so a correctly armed `backlog` entry left the guard printing byte-identical text and exiting 2, with no way out. That path honours the key it advises now, and reports one it has to ignore.
- **HIGH** — `edit_while_a_ready_pr_sits.py` and `branch_from_unmerged.py` both read `deferred()` and neither was wired, against this PR's own stated rationale that the readers move together. Both wired.
- **MEDIUM** — a stamp already past the expiry when it was written was silently dropped, and the reason it could be written at all was in the guards themselves: the advice printed a frozen `{int(time.time())}` literal, so copying it fifty minutes later produced exactly that entry. Reported now, and the advice is `$(date +%s)`.
- **MEDIUM** — the docstring claimed `None` meant "there is none to honour", when it also means live and honoured. Corrected: this function answers only "written, and rejected"; `deferred` is what distinguishes live from absent.
- **LOW** — `stamp.isdigit()` is true of characters `int()` rejects (`²`), and the new call site runs first, so `unsettled_pr.py` raised `ValueError` and exited **1** — which Stop treats as non-blocking. A guard turning itself off is the worst outcome here, and it is why this one was fixed before the MEDIUMs. Both readers go through a strict parse.
- **LOW, not fixed** — nothing pins any of this, because `.claude/hooks/` has no test harness and `test.yml`'s push paths do not cover it, unlike every other Python tool in the repository. That is a real gap and larger than this PR; it is #574's neighbourhood and belongs with whatever harness that lands.

Six inputs driven against `unusable`: a future stamp, a non-numeric field, a superscript digit, a stale-on-arrival stamp, a live entry and an absent key — plus `deferred()` on the superscript input to confirm it no longer raises, and the real Stop guard driven end to end on the stale-on-arrival case.

s4k10503:
Review round 2. Both findings were mine and both are withdrawals rather than repairs.

- **HIGH** — round 1 added a "stale on arrival" report for a stamp already past the expiry when written, and that contradicted the docstring's own statement that expiry is not reported. The reviewer is right and the reason is sharper than a contradiction: **the file cannot carry that distinction.** A stamp says when the deferral was written and nothing says when the line was added, so an entry written already stale and one written fresh that has since expired are byte-identical. The report claimed a distinction the data does not have. Withdrawn, and the docstring now says why rather than only that.

  The defect that motivated it is still fixed: the guards printed a frozen `{int(time.time())}` literal, so copying the advice fifty minutes later produced exactly such an entry. That is `$(date +%s)` now, which is where the fix belonged.

- **MEDIUM** — round 1 made the unreadable-`gh` path honour `backlog`, because that is the key the path's own advice named. Honouring it was right; the advice was wrong. One line should not silence two independent guards — that is the unqualified exemption the whole expiry mechanism exists to prevent. The path has its own `pr-list` key now and advises that.

Four inputs re-driven against `unusable`: a future stamp and a garbage field reported; an expired entry and a live one both silent.

PR #588, opened 2026-08-09
tooling(velvet): pin every PreToolUse hook's tool set against its matcher

### LINKED ISSUE #574: HELD_TOOLS and the settings matcher are a hand-kept pair, and a trim to either is silent

s4k10503:
Closes #574.

A hook decides which tools it handles; `.claude/settings.json` decides which reach it. Nothing compared them, and both drift directions are silent: trim a tool from the hook's set and the matcher still routes it, so the hook returns 0 and the guard is off for that tool with the suite still green; trim it from the matcher and the set keeps claiming a tool the hook never sees.

The issue asked for a measurement first, and it corrected the issue's own generalisation. "Every Bash-matcher hook checks `tool_name`" is false — two of the eleven do not check at all and rely on the matcher. And the new CHANGELOG guard on `tooling/changelog-closed-section` admitted `("Edit", "Write")` behind an `Edit|Write|NotebookEdit` matcher, which is a live mismatch rather than a hypothetical one.

Every PreToolUse hook declares `HOOK_TOOLS` and gates on it, the same one-place shape as the existing `HOOK_SCOPE`. `HookWiringCoverageTests` asserts the declaration exists, equals its matcher's alternation, and is actually read by the hook — a declaration nothing reads is the same silence as no declaration.

## The gate probe's payload table is hand-kept, and the guard after this one has the same problem

The gate probe poses each declaring hook a payload under a tool it is routed and the same payload under a name nothing routes. The payloads are a table in the fixture, and a guard the table poses nothing about answers none of them — which the fixture reported as a gate admitting the opposite of what it declares.

That is not hypothetical. `DeclaringHooks()` scans `.claude/hooks/**/*.py` and enrols anything declaring `HOOK_TOOLS`, so `pr_body_of_another_branch.py` joins the moment both branches land. Measured against all seven payloads with `cwd` an empty directory, that guard exits 0 with both streams empty every time. Whichever of the two landed second would have reddened `HookWiringCoverageTests` on `main`, against a correct gate.

One row fixes this instance — `gh pr create --title x --body-file velvet-no-such-body.md`, which that guard answers with exit 2 and "does not exist" under `Bash` and is silent under the unrouted name. The general question is what happens to the guard after that one.

The answer is not another mechanism. The check already fails loudly when a guard answers nothing, so nothing here is silent; what was wrong is where the failure pointed. It named the guard, and described only one of the two things that produce it. It now writes out both readings — a gate that returns before its readings for a tool it is routed, and a `GatePayloads` table holding nothing that guard decides about — and says that a guard added since the table was last extended wants a row there rather than a change to itself. CONTRIBUTING states the same, beside the rest of what the fixture fails on.

Deriving the payloads instead of listing them is what would remove the table, and there is nothing to derive them from: what a guard decides about is the guard's own subject, and reading it back out of the guard is the same mirror problem one level down.

## Also in this round

- **What counts as an answer.** Exit 0 plus anything on either stream counted text written before `main()` reads `tool_name`. A module-level `warnings.warn` — or any import-time `DeprecationWarning` out of a hook or its `lib/` — made every payload answer, including the one posed under the unrouted name, and the probe then reported a correct gate as reading something other than the event's tool name. What a hook writes with no event to read is now measured once per hook and subtracted. Scoring stdout alone closes the same hole, and it silences `merge_onto_unpublished_release.py`, whose only answer under any payload is `published_check.py`'s except-path note on stderr.
- **A pipe-ordering deadlock in the runner.** `Answer` read stderr to EOF before stdout, with the timed wait after both, so a guard writing past the pipe buffer blocked with nothing left to time it out. Measured against a child writing 2 MB: blocked past 15 s in that order, and returning the whole 2 MB when both streams are drained together.
- **`DeclaredTools`' null return** was justified by a collision `RoutedTools` can no longer produce. The null is load-bearing for the other reading: an empty string is not null, and the declaration check would pass a hook whose gate admits no tool at all.
- **Two CONTRIBUTING corrections.** It described the unrouted-name branch as what catches an inverted gate; such a gate answers nothing under the tools it is routed, so it is caught before that branch is reached. And it called `*` a documented all-tools form, which is a claim about Claude Code's matcher semantics that nothing here pins. What is decided here is that a matcher naming no set a check can read is refused — `*`, an empty string, and a `PreToolUse` entry carrying no `"matcher"` key at all, which is the shape the `SessionStart`, `Stop` and `SubagentStop` entries take.

EditMode 3788/3788, PlayMode 161/161, `assert_no_inconclusive.py` clean on both. Merged with `tooling/pr-body-provenance` onto `main`: EditMode 3792/3792, clean.

### DIFF
--- a/.claude/hooks/refuse/background_relative_path.py
+++ b/.claude/hooks/refuse/background_relative_path.py
@@ -25,6 +25,9 @@ sys.path.insert(0, str(Path(__file__).resolve().parent.parent / "lib"))
 from shell_commands import unexpanded  # noqa: E402  (path set above)
 
 
+HOOK_TOOLS = {"Bash"}
+
+
 # Nothing here reads a path's contents: a token the shell will still rewrite may or may not be
 # relative, and a background command that gets it wrong fails silently, so the guard errs toward
 # allowing rather than refusing every command carrying a variable.
@@ -64,7 +67,7 @@ def main():
         event = json.load(sys.stdin)
     except Exception:
         return 0
-    if event.get("tool_name") != "Bash":
+    if event.get("tool_name") not in HOOK_TOOLS:
         return 0
 
     tool_input = event.get("tool_input", {})
--- a/.claude/hooks/refuse/blind_git_add.py
+++ b/.claude/hooks/refuse/blind_git_add.py
@@ -19,6 +19,9 @@ from pathlib import Path
 sys.path.insert(0, str(Path(__file__).resolve().parent.parent / "lib"))
 from shell_commands import git_invocations
 
+
+HOOK_TOOLS = {"Bash"}
+
 # The sweeping forms. `-u` is not among them: it stages tracked modifications and cannot pick up a
 # file that arrived by accident, which is what both incidents were.
 #
@@ -53,7 +56,7 @@ def main():
         event = json.load(sys.stdin)
     except Exception:
         return 0
-    if event.get("tool_name") != "Bash":
+    if event.get("tool_name") not in HOOK_TOOLS:
         return 0
     command = event.get("tool_input", {}).get("command", "")
     if not sweeps(command):
--- a/.claude/hooks/refuse/branch_from_unmerged.py
+++ b/.claude/hooks/refuse/branch_from_unmerged.py
@@ -23,6 +23,9 @@ from deferrals import deferred
 from shell_commands import command_segments, git_invocation, tokens_of, without_redirections
 from velvet_hooks import BRANCH_BASES
 
+
+HOOK_TOOLS = {"Bash"}
+
 CHECKOUT_CREATE = ("-b", "-B")
 SWITCH_CREATE = ("-c", "-C", "--create", "--force-create")
 
@@ -220,7 +223,7 @@ def main():
         event = json.load(sys.stdin)
     except Exception:
         return 0
-    if event.get("tool_name") != "Bash":
+    if event.get("tool_name") not in HOOK_TOOLS:
         return 0
 
     made = creations(event.get("tool_input", {}).get("command", ""))
--- a/.claude/hooks/refuse/commit_failing_fast_checks.py
+++ b/.claude/hooks/refuse/commit_failing_fast_checks.py
@@ -21,6 +21,9 @@ from pathlib import Path
 sys.path.insert(0, str(Path(__file__).resolve().parent.parent / "lib"))
 from shell_commands import git_invocations, unexpanded
 
+
+HOOK_TOOLS = {"Bash"}
+
 NEUTER_CUTS = "scripts/test_quality/neuter_cuts.json"
 
 # `git commit` options that take a value, so their argument is not mistaken for a pathspec.
@@ -287,7 +290,7 @@ def main():
         event = json.load(sys.stdin)
     except Exception:
         return 0
-    if event.get("tool_name") != "Bash":
+    if event.get("tool_name") not in HOOK_TOOLS:
         return 0
 
     commits = commit_invocations(event.get("tool_input", {}).get("command", ""))
--- a/.claude/hooks/refuse/edit_while_a_ready_pr_sits.py
+++ b/.claude/hooks/refuse/edit_while_a_ready_pr_sits.py
@@ -41,7 +41,7 @@ GRACE = 900
 # Two polls plus a margin: one missed poll is a slow API call, two is a watcher that stopped.
 HEARTBEAT_TTL = 180
 
-HELD_TOOLS = {"Edit", "Write", "NotebookEdit"}
+HOOK_TOOLS = {"Edit", "Write", "NotebookEdit"}
 
 
 def watcher_age():
@@ -76,7 +76,7 @@ def main():
         event = json.load(sys.stdin)
     except Exception:
         return 0
-    if event.get("tool_name") not in HELD_TOOLS:
+    if event.get("tool_name") not in HOOK_TOOLS:
         return 0
 
     age = watcher_age()
--- a/.claude/hooks/refuse/merge_branch_held_by_worktree.py
+++ b/.claude/hooks/refuse/merge_branch_held_by_worktree.py
@@ -25,6 +25,9 @@ from shell_commands import program_invocations, unexpanded
 import repository
 
 
+HOOK_TOOLS = {"Bash"}
+
+
 # An operand the shell has not expanded yet resolves to nothing readable, and a merge guard errs
 # toward refusing: allowing means --delete-branch half-fails after the merge has already landed.
 UNEXPANDED_POLICY = "refuse"
@@ -90,7 +93,7 @@ def main():
         event = json.load(sys.stdin)
     except Exception:
         return 0
-    if event.get("tool_name") != "Bash":
+    if event.get("tool_name") not in HOOK_TOOLS:
         return 0
 
     cwd = event.get("cwd") or "."
--- a/.claude/hooks/refuse/merge_onto_unpublished_release.py
+++ b/.claude/hooks/refuse/merge_onto_unpublished_release.py
@@ -18,6 +18,9 @@ sys.path.insert(0, str(Path(__file__).resolve().parents[3] / "scripts" / "releas
 import published_check
 
 
+HOOK_TOOLS = {"Bash"}
+
+
 # No operand takes part in the verdict: the base holds an unpublished release or it does not, whatever
 # pull request is named.
 UNEXPANDED_POLICY = "n/a"
@@ -33,7 +36,7 @@ def main():
         event = json.load(sys.stdin)
     except Exception:
         return 0
-    if event.get("tool_name") != "Bash":
+    if event.get("tool_name") not in HOOK_TOOLS:
         return 0
 
     command = event.get("tool_input", {}).get("command", "")
--- a/.claude/hooks/refuse/merge_unproven_head.py
+++ b/.claude/hooks/refuse/merge_unproven_head.py
@@ -30,6 +30,9 @@ sys.path.insert(0, str(Path(__file__).resolve().parent.parent / "lib"))
 from shell_commands import program_invocations, unexpanded
 import repository
 
+
+HOOK_TOOLS = {"Bash"}
+
 TERMINAL_PASS = frozenset({"pass", "skipping"})
 
 # An operand the shell has not expanded yet resolves to nothing readable, and a merge guard errs
@@ -91,7 +94,7 @@ def main():
         event = json.load(sys.stdin)
     except Exception:
         return 0
-    if event.get("tool_name") != "Bash":
+    if event.get("tool_name") not in HOOK_TOOLS:
         return 0
 
     cwd = event.get("cwd") or "."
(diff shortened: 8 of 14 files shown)


s4k10503:
## Review round: the tool-set checks now compare sets, and the gate is run

Each finding was reproduced against `b38cfbf` before being fixed, and each fix was re-measured with the
same defect planted.

**Two sides saying nothing compared equal.** An empty `"matcher"` beside `HOOK_TOOLS = {}` left both
sides as the empty string, so the check that exists to catch disagreement reported agreement while the
matcher routed every tool and the gate admitted none. Reproduced: all three tool-set tests green.
A declaration naming no tool now answers as no declaration, and a side that names no set is reported
rather than compared.

**A wildcard matcher had no expressible answer.** `"matcher": "*"` compared equal only to
`HOOK_TOOLS = {"*"}`, which no real tool is in, so the guard never ran. Reproduced: green.
Decision: refuse it rather than expand it. Expanding `*` means holding Claude Code's whole tool list
in the fixture, where nothing updates it when the product gains a tool — a mirror with no source. A
matcher is now required to be an alternation of plain tool names, so a guard covering several tools
names them, and an empty matcher is refused the same way. `CONTRIBUTING.md` states the rule.

**An inverted gate was invisible** — the one failure a mechanical refactor of twelve gate lines
produces. Dropping the `not` from `blind_git_add.py` left all three tests green, and the pre-existing
verdict test stays green because its failure condition accepts an exit of 0. A new test poses each
declaring guard a payload it answers, under every tool it declares and under a name nothing routes,
and requires only the first to be answered. Answering is read from exit 2 *or* anything written at
exit 0, because `blind_git_add.py` refuses with a `permissionDecision` of `deny` and exits 0. The
payloads are posed with a scratch directory and a scratch `HOME`, so no answer depends on which tree
the suite runs in or on what the developer's pull-request watcher last wrote.

**Prose satisfied the read check.** `File.ReadLines` strips nothing, so a comment naming `tool_name`
and `HOOK_TOOLS` on one line passed while the gate compared a literal. Reproduced: green. Comments and
triple-quoted spans are now dropped before the reading. String literals are kept — the gate's own
subject is the string `"tool_name"`, so a full strip would take the line being looked for.

**A hook under two matcher entries could not be green.** Registering `blind_git_add.py` under a second
entry and declaring the correct union was RED both ways:

```
.claude/hooks/refuse/blind_git_add.py: matcher routes Bash, HOOK_TOOLS admits Bash|Edit|Write
.claude/hooks/refuse/blind_git_add.py: matcher routes Edit|Write, HOOK_TOOLS admits Bash|Edit|Write
```

The entries on a hook are now unioned before the comparison; the same registration is green.

**A false sentence in `CONTRIBUTING.md`.** "so the two halves are one edit apart instead of two" was
wrong — before it was matcher plus the gate's literal, after it is matcher plus `HOOK_TOOLS`, still
two. Replaced with what the declaration actually buys: it leaves the two halves comparable.

### Evidence

With all four of the first defects planted against the fix, the failure text is:

```
Given_TheHooksDeclaringAToolSet_When_EachIsPosedUnderARoutedNameAndAnUnroutedOne_Then_OnlyTheRoutedOneAnswers
  .claude/hooks/refuse/background_relative_path.py answers nothing posed as Bash
  .claude/hooks/refuse/shared_git_state.py answers nothing posed as Bash
Given_TheHooksDeclaringAToolSet_When_TheirSourceIsRead_Then_EveryOneComparesToolNameAgainstIt
  .claude/hooks/refuse/shared_git_state.py declares HOOK_TOOLS and gates on something else
Given_ThePreToolUseRegistrations_When_EachHookIsRead_Then_EveryOneDeclaresItsToolSet
  .claude/hooks/refuse/metadata_less_create.py is routed ""
Given_ThePreToolUseRegistrations_When_EachMatcherIsComparedWithTheHooksOwnSet_Then_TheyNameTheSameTools
  .claude/hooks/refuse/metadata_less_create.py: matcher "" routes no set of tool names, HOOK_TOOLS admits nothing
  .claude/hooks/refuse/merge_without_branch_deletion.py: matcher "*" routes no set of tool names, HOOK_TOOLS admits *
```

The same four planted against `b38cfbf` produced 9 passed, 0 failed.

Suites on the fixed branch: EditMode 3788 passed / 0 failed / 0 inconclusive, PlayMode 161 passed /
0 failed / 0 inconclusive, `assert_no_inconclusive.py` clean on both.

No hook behaviour changed in this round, so no CHANGELOG entry: the whole change is the fixture and
the contributor guide.

s4k10503:
Review round: the cross-branch defect plus five findings on this branch.

**The cross-branch one.** `GatePayloads` is a hand-kept table and `DeclaringHooks()` enrols every hook declaring `HOOK_TOOLS` automatically, so `pr_body_of_another_branch.py` on `tooling/pr-body-provenance` joins this check the moment both land. Posed all seven payloads under `"tool_name":"Bash"` with `cwd` an empty temp directory it gives exit 0, empty stdout, empty stderr every time. Measured on a scratch merge of both branches onto `main`, unmodified:

```
Given_TheHooksDeclaringAToolSet_When_EachIsPosedUnderARoutedNameAndAnUnroutedOne_Then_OnlyTheRoutedOneAnswers
  a gate reaching its readings for the tools it is not routed, or returning before them for the
  tools it is, admits the opposite of what it declares:
.claude/hooks/refuse/pr_body_of_another_branch.py answers nothing posed as Bash
```

An eighth payload — `gh pr create --title x --body-file velvet-no-such-body.md` — answers it: exit 2 with "does not exist" under `Bash`, exit 0 and silent under `VelvetNoToolIsCalledThis`. Appended rather than inserted, so every existing guard keeps the payload it already answered first. Same merge with the fix: 13/13.

The general question and its answer are in the body above.

**What counts as an answer.** Scoring exit 0 plus anything on either stream counted text written before `main()` reads `tool_name`. RED, with a module-level `warnings.warn` above one guard's `HOOK_TOOLS` and the old scoring:

```
.claude/hooks/refuse/blind_git_add.py answers a merge naming an unexpanded pull request
posed as VelvetNoToolIsCalledThis, which nothing routes to it
```

— a correct gate, accused of not reading the tool name, because the warning answered under every name including the unrouted one. What a hook writes with no event to read is measured once per hook and subtracted; with the same warning still in place the fixture is 10/10.

The suggested alternative of scoring stdout alone would have closed the same hole and opened another: `merge_onto_unpublished_release.py`'s only answer under any of the payloads is a stderr note, and removing that one `print` from `published_check.py`'s except path takes it to exit 0 with both streams empty — "answers none of them" against an untouched, correct gate. That dependency is not removed here, only made legible: the failure message now names the payload table as one of its two readings. Making it impossible would need a purpose-built repository fixture for that one guard to reach a real verdict in, which is more machinery than the finding is worth. Flagged rather than fixed.

**The pipe-ordering deadlock.** `Answer` read stderr to EOF before stdout, and `WaitForExit(60000)` sat after both, so a guard writing past the pipe buffer blocked with nothing left to time it out. Measured against a child writing 2 MB to stdout and one byte to stderr: that ordering was still blocked after 15 s; drained together, it returns all 2 MB. Not pinned by a committed test — a test for it hangs rather than fails.

**`DeclaredTools`.** Its comment justified the null by a collision this same commit made unreachable: `RoutedTools` never returns `""`. The null is still load-bearing on the other reading — an empty string is not null, so the declaration check would pass a hook whose gate admits no tool at all.

**CONTRIBUTING.** Two corrections, both in the body above: the branch that catches an inverted gate, and the unpinned claim about `*`.

Full runs on this branch: EditMode 3788/3788, PlayMode 161/161, `assert_no_inconclusive.py` clean on both. Merged with `tooling/pr-body-provenance` onto `main`: EditMode 3792/3792, clean. No CHANGELOG entry — contributor tooling, no user-visible behaviour.

s4k10503:
Pushed 2d068ce.

**The MEDIUM is confirmed.** With `stale_merge.py`'s `try/except` around `json.load` removed, the fixture as it stood fails:

```
Given_TheHooksDeclaringAToolSet_When_EachIsPosedUnderARoutedNameAndAnUnroutedOne_Then_OnlyTheRoutedOneAnswers
  Expected: <empty>
  But was:  < ".claude/hooks/refuse/stale_merge.py answers a merge naming an unexpanded pull request
              posed as VelvetNoToolIsCalledThis, which nothing routes to it" >
```

That is a correct gate accused of reading something other than the tool name — the misdiagnosis the previous round removed. The traceback becomes the baseline, so the silence the hook keeps for every payload it *can* read differs from it and scores as an answer under every name.

Posed `{}` instead, the same removal leaves the fixture green: 10 passed, 0 failed with the parse guard still gone. Both baselines report 0 faults on the unmutated branch, so the change is behaviour-preserving today.

**Both LOWs are confirmed and fixed.**

- `LoadingOutput` is cleared in `[OneTimeSetUp]` beside the scratch directory it is measured under.
- The third cause of "answers none of them" is real and now written in both places: making a hook raise gives `stale_merge.py answers none of the 8 payloads posed as Bash`, which is neither an inverted gate nor a missing row. `CONTRIBUTING.md` now names `GatePayloads` as the table a new guard wants a row in.

EditMode: 3788 passed, 0 failed, 0 inconclusive, 0 skipped; `assert_no_inconclusive.py` clean. No CHANGELOG entry — contributor tooling. `CONTRIBUTING.md` is the guide this touches and is updated in the same commit.

s4k10503:
Follow-up 6296efd, from the sweep over my own prose.

The reason I first wrote beside the cache clear — "a hook reading its state out of HOME writes something else under the new one" — was an unverified mechanism: both scratch directories are fresh and empty, so the one hook that reads HOME answers the same under either. Replaced with the constraint that actually holds: a baseline is comparable only with answers measured under the same HOME, and that directory is replaced there. The assertion string is rewrapped in the same commit; the text it builds is byte-identical.

EditMode re-run on the pushed head: 3788 passed, 0 failed, 0 inconclusive, 0 skipped, `assert_no_inconclusive.py` clean. PlayMode: 161 passed, 0 failed, 0 inconclusive.

Issue #432, opened 2026-08-06
V.TextField models no multiline, mask or mobile-keyboard knob, and takes no event or onCreated parameter


s4k10503:
> **Re-measured against `origin/main` (`89042a0`) on 2026-08-13.** Four of the ten knobs this issue
> asked for shipped in #589, and the pool-reset half it closed with was separately closed by #576.
> The original body's measurements are superseded where marked; what remains is below.

## What uGUI does

InputField / TMP_InputField carry characterLimit, lineType (single-line / multi-line), contentType (integer, decimal, email, password, alphanumeric…), placeholder text, readOnly, caret and selection colour, onEndEdit / onSubmit (commit-on-blur-or-Enter, distinct from per-keystroke change), and the mobile keyboard knobs (keyboard type, autocorrection, hide-mobile-input).

## What Velvet does

`V.TextField(className, value, onValueChanged, key, name, label, isPasswordField, placeholder, maxLength, isReadOnly, isDelayed, enabled, refCallback, while*Class, data, aria)`.
`TextFieldSettings` is a five-field record — `IsPassword`, `Placeholder`, `MaxLength`, `IsReadOnly`,
`IsDelayed` — and `FiberPropApplier.ApplyTextField` writes all five, each undeclared-when-null against
a recorded constructed default.

Nothing else about the field is modelled: no `multiline`, `maskChar`, `selectAllOnFocus`,
`selectionColor`, `cursorColor`, `keyboardType`, `autoCorrection`, `hideMobileInput`,
`hideSoftKeyboard` or `verticalScrollerVisibility`. And unlike `V.ScrollView` and `V.Slider`,
`V.TextField` still does not take an `onCreated:` parameter, nor any submit / key / blur handler —
its only event is `onValueChanged`.

## What an author hits

A placeholder, a character limit, a read-only display field and a commit-on-Enter field are now
declarative. A multi-line chat/notes box, a PIN pad that raises a numeric soft keyboard on mobile,
and a custom mask character still require dropping into `refCallback` and writing UI Toolkit property
assignments by hand. None of it participates in the diff, so none of it can change with state. Submit
and blur handling is reachable only off the beaten path.

## Where

Packages/com.velvet.core/Runtime/Component/V.cs:561-596 (`V.TextField`);
Runtime/Reconciler/FiberElementProps.cs:159-164 (`TextFieldSettings`);
Runtime/Reconciler/FiberPropApplier.cs:216-239 (`ApplyTextField`).
[...]

PR #589, opened 2026-08-09
feat(velvet): declare a text field's placeholder, length limit, read-only and delayed flags

s4k10503:
Closes part of #432 — the four knobs that map one-to-one onto HTML input attributes. The TextArea / `inputMode` / extra-events half is still a design question and stays in the issue.

`V.TextField` exposed one knob of UI Toolkit's text-input surface. `placeholder`, `maxLength`, `readOnly` and `isDelayed` are the four with no naming ambiguity, and each is reachable today only by dropping to `V.Custom<TextField>` and mutating the element.

A prop a later render stops declaring restores what the element was constructed with rather than a constant, through the `ConditionalWeakTable` recorded-default shape `ApplyFocusable` established — the defect PR #578 fixed, not repeated.

No pool change was needed: all four are already scrubbed by `FiberPrimitiveElementPool`'s TextField reset, which PR #576 derived from the whole writable surface rather than from a list of props.

EditMode 3787/3787.

### DIFF
--- a/Packages/com.velvet.core/CHANGELOG.md
+++ b/Packages/com.velvet.core/CHANGELOG.md
@@ -10,8 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `V.TextField` declares four more of the text-input surface: `placeholder:`, `maxLength:`,
-  `isReadOnly:` and `isDelayed:` — HTML's `placeholder`, `maxlength` and `readonly`, plus the
-  commit-on-Enter-or-blur behaviour. Reaching them previously meant writing the UI Toolkit properties
+  `isReadOnly:` and `isDelayed:` — HTML's `placeholder`, `maxlength` and `readonly`, plus a value that
+  catches up with the typed text on Enter, on the field losing focus, and on a render taking
+  `isDelayed:` off. That last one reports through `onValueChanged:`, so turning the flag off mid-edit
+  hands the component the pending text instead of stranding it on screen.
+  Reaching them previously meant writing the UI Toolkit properties
   by hand from `refCallback:`, which left them outside the diff and so unable to change with state.
   Each is undeclared when null: a member no render has declared is left alone, and one a render
   declared and a later one dropped restores the value the field carried before any render declared it.
--- a/Packages/com.velvet.core/CHANGELOG.md
+++ b/Packages/com.velvet.core/CHANGELOG.md
@@ -18,16 +18,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- A `TextField`'s `isPasswordField` was written back to `false` by any render that declared another
-  text-input prop, so a mask a `refCallback:` had switched on came off on the next render that changed
-  the field's placeholder, length limit, read-only or delayed flag. A member no render declares is now
-  nobody's to write.
-
-- A pooled widget carried the props a previous consumer had declared on it into its next mount: the
-  recorded value a dropped `Focusable`, `TabIndex`, `DelegatesFocus` or text-input prop restores
-  belongs to the tree that declared it, and survived into the tree that recycled the element. The
-  next consumer's own assignment to the same property was then overwritten on the first re-render
-  that dropped a prop or changed a neighbouring one.
+- A pooled widget carried the `Focusable` default recorded under a previous consumer into its next
+  mount, so a render that dropped a declared `Focusable` prop handed the element that consumer's
+  focusability rather than the one it carried when this consumer first declared the prop. The record
+  is forgotten on the pool return every poolable type passes through.
 
 - A `TabIndex` or `DelegatesFocus` prop that a later render stopped declaring was written back as `0`
   or `false` rather than as the value the element was constructed with — neither of which is the
--- a/Packages/com.velvet.core/CHANGELOG.md
+++ b/Packages/com.velvet.core/CHANGELOG.md
@@ -14,9 +14,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   commit-on-Enter-or-blur behaviour. Reaching them previously meant writing the UI Toolkit properties
   by hand from `refCallback:`, which left them outside the diff and so unable to change with state.
   Each is undeclared when null: a member no render has declared is left alone, and one a render
-  declared and a later one dropped restores what the element was constructed with.
+  declared and a later one dropped restores the value the field carried before any render declared it.
+  The same four arrive on `Velvet.Experimental.VTextField` as `Placeholder`, `MaxLength`, `IsReadOnly`
+  and `IsDelayed`.
+
+### Changed
+
+- `V.TextField`'s four new parameters sit between `isPasswordField:` and `enabled:`, so a call passing
+  arguments positionally past `isPasswordField` now binds them to different parameters. Every argument
+  there but a `null` literal changes type across the shift, so such a call fails to compile rather than
+  rebinding silently; pass them by name.
 
 ### Fixed
+
+- A `TextField`'s `isPasswordField` was written back to `false` by any render that stopped declaring
+  it, so a mask a `refCallback:` had switched on came off on the first render that dropped the prop.
+  A member a render declared and a later one dropped now restores what the field carried when the
+  prop was first declared, and one no render has declared is nobody's to write.
 
 - A pooled widget carried the `Focusable` default recorded under a previous consumer into its next
   mount, so a render that dropped a declared `Focusable` prop handed the element that consumer's
--- a/Packages/com.velvet.core/CHANGELOG.md
+++ b/Packages/com.velvet.core/CHANGELOG.md
@@ -13,10 +13,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `isReadOnly:` and `isDelayed:` — HTML's `placeholder`, `maxlength` and `readonly`, plus the
   commit-on-Enter-or-blur behaviour. Reaching them previously meant writing the UI Toolkit properties
   by hand from `refCallback:`, which left them outside the diff and so unable to change with state.
-  Each is undeclared when null and restores what the element was constructed with once a later render
-  stops declaring it.
+  Each is undeclared when null: a member no render has declared is left alone, and one a render
+  declared and a later one dropped restores what the element was constructed with.
 
 ### Fixed
+
+- A `TextField`'s `isPasswordField` was written back to `false` by any render that declared another
+  text-input prop, so a mask a `refCallback:` had switched on came off as soon as the field's
+  placeholder or label changed. A member no render declares is now nobody's to write.
+
+- A pooled widget carried the props a previous consumer had declared on it into its next mount: the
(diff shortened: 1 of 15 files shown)


s4k10503:
Review round 1. The MEDIUM was a real defect and fixing it surfaced a second one.

- **MEDIUM** — `ApplyTextField` held **one record for five members**, so once any one of them had been declared, every later call wrote all five. A member a render never declared was clobbered. The comment claiming "same recorded-default shape as `ApplyFocusable`" was what hid it: `ApplyFocusable` holds one record **per prop** and writes an undeclared prop only when that prop has a record. It is per-member now, split into five small appliers — keeping it inline compiled at 19 branching decisions against a limit of 20, one edit from the wall.

  `isPasswordField` is fixed in the same mechanism rather than filed. It is one branch of the same shape, and leaving it would have restored the two-shape file the finding named. It pinned trivially: same sequence, `Expected: (True, True) But was: (True, False)`.

- **A second defect, found while fixing.** The recorded defaults **survived a pool cycle**. `VNodePool` is process-global and LIFO and `Reconciler.Dispose` does not empty it, so a field that recorded a placeholder in one tree came back out carrying that record; the next tree's `refCallback` writes its own placeholder and the first re-render restores the *previous tenancy's* value over it. The same leak exists for `Focusable` / `TabIndex` / `DelegatesFocus` from #578, already on main, through a narrower door. One call at `ReturnToPool` — the only method in the package that returns anything to those pools — covers all four tables.

  This does not contradict the review's "pool ghosting is unreachable" note; that note is right about the properties. The reset makes a stale **record** invisible rather than absent, because restoring a value equal to the reset value is a no-op until something writes in between. The property leak and the record leak are different objects.

- **LOW** — two test names asserted engine behaviour the tests do not observe. Renamed to what is measured — and two more in the same fixture (`Then_the_element_shows_it`, `..._shows_no_hint`) are the same class of claim, since the assertions read `textEdition.placeholder` and observe no rendering. Renamed too. Naming brought to per-segment PascalCase.

Nine new tests, three separate neuters each reverted after measuring: the pre-fix one-record shape (6 red), an emptied `ForgetRecordedDefaults` (4 red), and the `ReturnToPool` call site alone (1 red — the emptied body alone would not have pinned the call site).

EditMode 3803/3803, PlayMode 161/161, Generators~ 422/422, `assert_no_inconclusive.py` clean on both runs.

s4k10503:
### Review round: two findings taken, one refuted

**MEDIUM — `FiberPropApplier.ForgetRecordedDefaults` comment. Confirmed and rewritten.**

The old sentence located the harm in the case where the next tree declares none of these props. Verified that case is inert: `FiberElementFactory.ApplyProps` guards each of them on `HasValue` / `props.TextField != null`, and `FiberNodePatcher.DiffProps` on value inequality, so with the prop undeclared on both sides no applier that reads a record runs at all.

The mechanism that is load-bearing is the one the pool test drives: a record is the applier's claim on a member, recording is idempotent, so a record left from the previous tenancy is not replaced by the next tenancy's first declared write — and the restore then puts the previous tenancy's reading over whatever this one wrote. Rewritten to that.

**MEDIUM — CHANGELOG "or label". Confirmed and corrected.**

`label` lands in `props.Text` (`V.cs:586`), the five text-input props in `props.TextField` (`:588`). `TextFieldSettings` is a `sealed record`, so `oldProps.TextField != newProps.TextField` is structural and two renders differing only in the label skip `ApplyTextField`. The sentence now names the four props that do trigger it.

**LOW — obsolete forwarders. Refuted; no change.**

The `[Obsolete("… Use textEdition.… instead.")]` forwarders for `isReadOnly`, `maxLength`, `maskChar` and `isPasswordField` are declared on the **nested `TextInputBaseField<T>.TextInputBase`** class, not on `TextInputBaseField<T>` itself. The applier writes `field.isPasswordField` / `field.maxLength` / `field.isReadOnly` where `field` is a `TextField`, which binds to the outer `[CreateProperty]` properties — none of which carries `[Obsolete]`. Measured on `6000.3.11f1`: the EditMode compile emits 44 `CS0618` warnings across the project and **zero of them are in `FiberPropApplier.cs`** (they are in `FiberElementPoolReset.cs` and `FiberPrimitiveElementPool.cs`, for `cacheAsBitmap` and `ITextSelection.cursorColor` / `selectionColor`).

`ApplyPlaceholder` uses `textEdition.placeholder` not as a style choice but because the outer `TextInputBaseField<T>.placeholderText` is `internal` — `textEdition` is the only public route. So there is no inconsistency to resolve.

Separately, the previous round's stated reason for the public `maxLength` setter needed correcting. Measured: setting `value = "abcdefgh"` then narrowing the limit to 3 leaves `field.value` at `"abcdefgh"` under **both** spellings, and clips the displayed text to `"abc"` under **both**. The difference is on widening — only `TextField.maxLength`'s setter re-derives the displayed text from the field's value, so only it puts the clipped characters back. Restoring a dropped limit is such a widening. That is now pinned by a test rather than left as prose.

### RED evidence

`Given_AMaxLengthThatClippedTheValue_When_ALaterRenderDropsIt_Then_TheClippedCharactersComeBack`, with `ApplyMaxLength` rewritten to the `textEdition.maxLength` spelling:

```
Expected: (abc, abcdefgh)
But was:  (abc, abc)
```

Reverted, it passes. The `abc` term is folded into the assertion because the restored reading alone is also what an implementation that never clipped would produce.

### Suites

EditMode 3804 passed / 0 failed / 0 inconclusive / 0 skipped. PlayMode 161 passed / 0 failed / 0 inconclusive / 0 skipped. `assert_no_inconclusive.py` clean on both. No `error CS` in either log.

### Docs and CHANGELOG

`Documentation~/react-migration.md` already carries the four props and needs no further change — this round altered no behaviour. No new CHANGELOG entry: the comment fix and the test are not user-visible, and the one CHANGELOG change is a correction to an entry this PR already added.

### Found, not fixed

- The project compiles with 44 `CS0618` warnings and nothing prevents more from appearing. Baselining that would need its own change.
- `V.TextField(value: "abcdefgh", maxLength: 3)` leaves `field.value` at 8 characters while the field displays 3. This is UI Toolkit's own behaviour under both spellings, not something this PR introduces, but the `maxLength:` parameter doc says "Maximum number of characters the field accepts", which reads stronger than what the value actually holds. HTML's `maxlength` constrains typing, not a programmatically set value, so this may be correct parity — flagging rather than changing it.

s4k10503:
Pushed 4aac71c.

**The MEDIUM is confirmed against the released tag, not just `main`.** At `v2.1.0-main`:

- `FiberElementProps.cs:154` — `public sealed record TextFieldSettings(bool? IsPassword = null);`, no other text-input member;
- `V.cs:577` — `props.TextField = isPasswordField.HasValue ? new TextFieldSettings(isPasswordField) : null;`
- `FiberNodePatcher.cs:1455` — `if (oldProps.TextField != newProps.TextField)` guards the only `ApplyTextField` call site.

So a tree that never declares `isPasswordField` leaves both sides null, the applier is never reached, and a mask set from `refCallback:` survives. The placeholder, length limit, read-only and delayed flags the entry named do not exist in any released version to change. Both the defect and its fix are this branch's own (`d9bbe05` then `47be613`). The entry is deleted rather than reworded: a defect introduced and fixed inside one unreleased version needs no `Fixed` entry.

**Checking the branch's other entries found one more of the same shape, partly.** The pooled-record entry named four props, and only one of them is released: `s_focusableDefaults` shipped in `2.0.1` (`git tag --contains` lists `release/2.0.1`, `v2.0.1-main`, `v2.1.0-main`), while `s_tabIndexDefaults` / `s_delegatesFocusDefaults` came from `34967e9`, which no release tag contains, and the text-input records are this branch's. A pooled widget carrying a stale `Focusable` record *is* a defect a released version shows, so the entry stays — narrowed to that.

The `Added` entry is sound: `V.TextField`'s four new parameters are user-visible and released in nothing.

EditMode: 3804 passed, 0 failed, 0 inconclusive, 0 skipped; `assert_no_inconclusive.py` clean. The change is `CHANGELOG.md` only, so there is no RED test to show — the evidence is the code reading above. No `Documentation~` claim referenced `isPasswordField`, so nothing else needed correcting.

s4k10503:
PlayMode on the pushed head: 161 passed, 0 failed, 0 inconclusive, 0 skipped.

s4k10503:
Review findings addressed in `ffb4726`.

**A stranded edit when `isDelayed` comes off mid-edit.** Making the flag a diffed prop opened a way to lose typed text. While the flag is set the field holds the text on its inner element without writing it to `value`, and the commit that would have written it (`TextField.HandleEventBubbleUp`, on `FocusOutEvent`) runs only while the flag is still set. So: type into a field rendered `isDelayed: true` → any re-render turns the flag off → blur commits nothing. The text stays on screen, `onValueChanged` never fires, and a render repeating the same `FieldValue` skips `ApplyFieldValue`, so nothing re-syncs it either.

`ApplyDelayedFlag` now routes both writes through a helper that commits `value = text` before clearing the flag. Committing rather than suppressing is what a controlled field needs — without the notification the parent's state stays stale and the next render writes it back over the typed text. The commit is the same one the engine performs on blur, and it dispatches through `BaseField.value`, so it queues rather than re-entering the flush.

Two cases in `TextFieldInputPropTests` cover both routes off the flag (a render redeclaring it `false`, a render dropping it). Both fail without the change with `Expected: "typed" / But was: <string.Empty>`.

**`focus.md`'s third focus-default category was false.** The guide said a UI Toolkit composite field mounted through `V.Custom` is built neither out of the tab ring nor delegating focus. `BaseField<T>`'s constructor writes `focusable = true; tabIndex = 0; excludeFromFocusRing = true; delegatesFocus = true`, and `Foldout` and `RadioButtonGroup` set `delegatesFocus = true` as well — no composite field is built doing neither, and nothing pinned the claim. The clause is deleted; the two categories `ConstructedFocusDefaultTests` does pin (`Label` at -1, `TextField` delegating) carry the argument on their own. The sentence arrived with #578 and is already on `main`, so this also removes it there.

**The `isPasswordField` entry, restored narrowed.** The neighbour scenario the deleted entry described is indeed unreleasable, but the drop scenario is not. On 2.1.0, render 1 `V.TextField(refCallback:)` with the callback setting `isPasswordField = true`, render 2 `V.TextField(isPasswordField: true, refCallback:)`, render 3 `V.TextField(refCallback:)` — render 3 leaves `newProps` null, `DiffProps` substitutes `Empty`, and `ApplyTextField(el, null)` writes `false`, taking the mask off. The entry now states that half only.

**An owner for the four props' null contract.** `react-migration.md` §2-1 now states it once for `placeholder:` / `maxLength:` / `isReadOnly:` / `isDelayed:` — null is not `placeholder=""`, not `maxLength: -1`, not `isReadOnly: false` — and links to `focus.md` for the same rule on the three focus props. `V.TextField`'s XML doc points at it instead of documenting it on `placeholder` alone.

**CHANGELOG completeness.** The `Added` entry now names `Velvet.Experimental.VTextField`'s four new properties, and a `Changed` line records that the four parameters sit between `isPasswordField:` and `enabled:`. Every argument in the shifted positions but a `null` literal changes type across the shift, so a positional caller past `isPasswordField` gets a compile error rather than a silent rebinding.

EditMode 3806 passed / 0 failed / 0 inconclusive / 0 skipped; PlayMode 161 / 0 / 0 / 0; `assert_no_inconclusive.py` clean on both; `test_release_notes.py` 26/26.

PR #593, opened 2026-08-10
fix(velvet): let a live registry portal follow a re-registered id

### LINKED ISSUE #412: A live registry portal never re-resolves its target: re-registering the id leaves children patching a destroyed element

s4k10503:
Closes #412.

A portal resolved `V.Portal("modal-root", …)` once at mount and then held the element, so
re-registering the id moved only future portals. A `"modal-root"` the UI itself owns — torn down on
navigation and re-registered from the rebuilt screen's `refCallback` — left every live portal's
children writing into the element that had been replaced, with no warning and no way back short of
changing the id or the key.

The children now follow the id, as an unmount and a remount: the range leaves the old element and is
created into the new one, which is what `createPortal` does for a changed container and the route
#581 took for `V.Portal(container:)`. State, refs and effects under the portal do not survive the
move. Re-registering the same element, and unregistering the id, still leave a live portal exactly
where it is — an unregistered id names nothing to move to.

## Reparent was not a real alternative

A portal's slot range addresses positions in the element it mounted into, and reusing it against
another element's children would diff one portal's range against another's — the objection the old
code's own comment raised, and the one React answers by deleting the portal fiber rather than
patching it.

## Where the swap is observed

`ReconcileKeying.CanPatch` cannot see it. Both nodes carry the same id string; what differs is the
element behind it, which no VNode holds, so the element-valued form's `ReferenceEquals` term has no
counterpart here. `FiberNodePatcher.ResolvePortalTarget` compares the registry against the element
the portal recorded instead, empties the old range through `FiberElementCleaner`, and falls into the
same tail a mount-before-registration heals through — which is also what attaches the
synthetic-bubbling bridge on the new element. The old element's bridge is released only when no live
portal still resolves to it, and the ranges of portals that sat after the departing one collapse left.

## The registration has to say so

Measured, not assumed: with the patch-time comparison alone and no signal from `Register`, the
children stay put. Nothing in a render reads the registry after a portal resolves its id, so a
re-registration from a `refCallback` in an unrelated subtree reaches no patch at all. `Register` now
asks the components that declared the portals on that id to render again. Dropping their memo cache
first is load-bearing on its own: a component whose hook inputs are unchanged is served the cached
VNode tree, and the reconciler skips a reference-identical node without patching it. A fixture whose
host has no hook cannot show that, because the weaver leaves such a body unwoven — the auto-memoized
host is the case that goes red without the invalidation.

The signal fires on every registration rather than only on an overwrite: a screen that unregisters as
it tears down leaves the table with nothing to compare against, while the portal still holds an
element.

## Verification

`PortalRegistryRetargetTests` covers the move, its remount identity, the append after the
replacement's own children, the downstream range on the element left behind, the bridge released
there and the bridge kept when another portal still resolves to it, and three host shapes for the
registration signal. Every case is red under a cut aimed at one term: the registry comparison (7
red), the `Register` signal (3), the memo drop (1), the slot start on the replacement (1), the
downstream shift (1), the bridge release (1), the release guard (1).

`PortalLayersTests.Given_ATargetReRegisteredMidLife_When_ThePortalPatches_Then_ItKeepsItsMountedTarget`
pinned the old rule and is replaced by the growth case in the new fixture.

EditMode 3839/3839, PlayMode 161/161, `assert_no_inconclusive.py` clean, generator suite 422/422.

`Documentation~/portals.md` states the new rule, replacing the paragraph that recorded the asymmetry
as deliberate; the `Documentation~/README.md` index row describes the guide's scope, which is
unchanged. The `[Unreleased]` CHANGELOG entry for the element-valued form drops the sentence this
change makes false, and a `Fixed` entry is added.

## Not fixed here

A portal that mounted while its id was unregistered still waits for an unrelated re-render to heal;
the registration signal deliberately skips those entries.


s4k10503:
Review round 1. Three HIGH and one MEDIUM, all confirmed and fixed — with two corrections to the findings themselves.

- **HIGH 1 (heal consumes the target's own children)** — confirmed, but **the second half of the stated mechanism is wrong**. `PatchPortalChildren` derives `newSlotLength = Count(target) - (beforeTailCount - prevState.SlotLength)` = 0, because the reconcile *patched* rather than inserted and the child count never grew. So the portal records a **zero-length** range: the unmount removes nothing and the container is left holding its own element carrying the portal's content. A leak, not a deletion. Also, `PatchOrReplaceAtSlot` does not remove a mismatched `Label` outright — `CanPatch` compares two VNodes (both `V.Div` here) and never looks at what element occupies the slot, so the `Label` is patched *as if* it were a `Div`.
- **HIGH 2** — confirmed as stated. `refCallback` fires synchronously inside `CreateElement`, so the sibling's registration lands during the same mount pass that just recorded `Target = null`.
- **HIGH 3** — confirmed, and **the disposal is fixed rather than the sentence corrected**, so the CHANGELOG, `portals.md` and the `ResolvePortalTarget` comment all become true. Selection is by slot-range intersection: it cannot be by containment (`IsInsideAny` walks *up*, and the target is the fiber's `MountPoint`), and it cannot be by `MountPoint` alone (one target carries several ranges).
- **MEDIUM** — the comment's mechanism did not exist. I did **not** replace it with the `CleanupElement` → `DisposeFibersUnder` one suggested, because that has the same `parentFiber`-mismatch doubt and asserting it would be a second unverified "X because Y". It states the ordering constraint with no mechanism claim.

One claim from the review I could not verify and therefore wrote nowhere: that the stale fiber is *handed back* on the next mount under the same key. `FiberStack.Current` differs between the drain mount and the heal patch, so the lookup plausibly misses and the fiber merely leaks. The measurement shows only that the cleanup never ran, and that is all the comments say.

**Evidence discipline.** Three fixes landed together, so a joint RED proves nothing about which one matters. Each of the five new tests was measured red at `afb41e1` **and** isolated by a single-change cut:

| cut applied alone | tests that went red |
|---|---|
| remove `DisposeInlineFibersInSlotRange` | exactly the two `ItsEffectCleanupRuns` |
| restore `info.Target == null` | exactly the sibling-refCallback case |
| remove the tail rebase | both heal tests **and** this PR's existing `Given_AReplacementHoldingItsOwnChildren_…` |

That last row is the point of the refactor: one expression guards both entrances now, so the retarget test and the late-resolution test go red at the same cut.

A universals sweep caught one of my own sentences: I had written that a portal waiting on an unregistered id "has nothing but this signal to wake it", which `PortalLayersTests.Given_ATargetRegisteredAfterMount_When_Patched_Then_TheMountHeals` contradicts.

EditMode 3844/3844, PlayMode 161/161, Generators 422/422, `assert_no_inconclusive.py` clean on both.

s4k10503:
### The defect, confirmed

`DisposeInlineFibersInSlotRange` selected by `ComponentFiber.MountSlotStart`, and nothing rewrites that field when a **neighbouring** portal's range on the same target moves. `PortalSlotTracker.ShiftSlotStartsAfter` rewrites `PortalSlotInfo.SlotStart` alone, so the two coordinates diverge and the selection read the one that is not maintained.

Two portals on one container is what `portals.md` documents and what an overlay holding a toast and a modal is. Both directions reproduce, and every case the branch already carried drives a single portal, which is why neither showed.

### RED on the parent (`38d245c` + `origin/main`)

Three new cases in `PortalRegistryRetargetTests`, run against the parent runtime with only the test file present:

```
FAIL: Given_ANeighbouringPortalThatGrewFirst_When_ItCloses_Then_TheOtherPortalsComponentChildSurvives
   Expected: 0
   But was:  1
FAIL: Given_ANeighbouringPortalThatClosedFirst_When_TheSecondCloses_Then_ItsComponentChildIsDisposed
   Expected: 1
   But was:  0
FAIL: Given_APortalWhoseOnlyChildRendersNothing_When_ItUnmounts_Then_ThatChildIsStillDisposed
   Expected: 1
   But was:  0
```

The first is the disposal of the wrong fiber: the modal's top-level component is torn down while the modal is open. The second is the mirror leak — the modal's own close no longer reaches its component, which is the leak this branch set out to close. The third is the declared narrowing, reachable from an ordinary shape: a component currently rendering `V.Fragment(Array.Empty<VNode>())` leaves the portal occupying no slots, so the `slotLength <= 0` guard returned and nothing was disposed.

Each arrangement is load-bearing: drop the growth step from the first and it passes on the parent; drop the earlier close from the second and it passes too.

### The fix

An inline fiber is stamped, at registration, with the portal placeholder whose children reconcile mounted it (`ComponentFiber.OwningPortalPlaceholder`, set from `ReconcilerContext.CurrentPortalPlaceholder`, which is set-and-restored at the deferred mount and at every patch — world-space and heal both funnel into `PatchPortalChildren`). The teardown disposes by that placeholder.

The selection no longer names a position, so both narrowings go with it: the zero-slot portal above, and a trailing child rendering nothing that landed on the range's exclusive end and was excluded while the same component at the head of the list was selected.

`CHANGELOG.md` and the comment both asserted a completeness the range selection did not have. Both now describe what the placeholder selection does; the CHANGELOG bullet is amended in place rather than joined by a new one, since the multi-portal defect never shipped.

### Suites

| | passed | failed | inconclusive |
|---|---|---|---|
| EditMode (full) | 3869 | 0 | 0 |
| PlayMode (full) | 161 | 0 | 0 |
| `Generators~` | 422 | 0 | 0 |

`assert_no_inconclusive.py` clean on both XMLs. No `-nographics`. `BundledStyleSheetInclusionTests` passed in this worktree — issue #596 appears to be gone since #595 landed.

### Found, not fixed — same divergence, older than this branch

`MountSlotStart` also goes stale for reads other than the teardown's. After the toast portal grows, an **isolated re-render** of the modal's portal component child patches at the slot the toast now occupies:

```
Expected: "toast-one|toast-two|content"   (before the modal's own state change)
But was:  "toast-one|changed|content"
```

It renamed the toast's second child and left its own untouched. This is not the branch's doing: `PortalSlotTracker` and both its call sites are byte-identical on `origin/main`. It wants a separate fix — maintaining `MountSlotStart` alongside `PortalSlotInfo.SlotStart` interacts with `FiberCommitWork.PropagateInlineSlotShift`, which walks the fiber-tree sibling chain and, at the portal drain, can put two different portals' children in one chain. I did not measure that interaction, so I did not touch it.

`origin/main` is merged in. `portals.md` needs no change: it already documents several portals sharing a container and never described the disposal selection.

s4k10503:
Review round 2. Both findings confirmed by measurement, and both corrected in prose rather than code — the design survives; only its justification was wrong.

**Finding 1 — right, and for a reason neither the review nor I had exactly.** `PropagateInlineSlotShift` does walk `fiber.Sibling` with no `MountPoint` check, and both portal children are siblings on the drain anchor, so a neighbour's growth **does** rewrite `MountSlotStart`:

```
a=0->0 b=1->2 sameParent=True aSiblingIsB=True bothMountOnTarget=True targetChildren=3
```

So "nothing rewrites it" is false. What actually leaves it stale is the *other* route — a neighbour whose own VNode child list changes:

```
m=1->1->1  treeAfterGrow=VisualElement/t1/t2/plain  treeAfterClose=VisualElement/plain
```

The element sits at slot 1, then 2, then 0, and `MountSlotStart` stays 1. That is the sentence the comment carries now.

**Finding 2 — the premise holds, the predicted consequence does not.** The key is identical on both sides (`portalCParent == hostFiber == movedCParent`, same explicit key, same identity), so "misses the lookup" was not why a new fiber appears. But the stale-stamp disposal does **not** reproduce, because the fiber does not survive the move at all:

```
stampedInPortal=True sameFiber=False stampAfterMove=False
cleanupsAfterMove=1 disposedAfterMove=True cleanupsAfterClose=1 disposedAfterClose=False
```

Taking option (b): a true comment plus a case pinning the limit. That case is green both ways and is **not** claimed as RED evidence — it is a guard that goes red if a re-home ever starts carrying the fiber across.

**Finding 3, new, and it is a real gap the commit claimed to have retired.** `[V.Div, V.Component(rendersNothing, key:"c")]` puts the component at `MountSlotStart == slotStart + slotLength`, the range's exclusive end, so `>= slotEnd` excluded it — while the same component written *first* is inside the range. Now pinned and RED at `Expected: 1 But was: 0`.

Four cases RED against the pre-branch runtime with the test file kept; 30/30 with the files restored. EditMode 3871/3871, PlayMode 161/161, `assert_no_inconclusive.py` clean on both.

**Three defects found and not fixed, all measured pre-existing** — cross-target `MountSlotStart` corruption (#602), a keyed component re-homed out of a portal losing its fiber with a one-render unowned window (#601), and a sibling portal closing remounting a component that lives in *another* portal (`setups 2→3, cleanups 1→2`). Every "pre-existing" claim was measured with this branch's runtime files replaced by their pre-branch contents **in the same worktree**, same reading, and `git diff origin/main HEAD -- FiberCommitWork.cs` is empty. This branch makes an existing shape easier to reach; it does not introduce it.

s4k10503:
Review round 3. The HIGH reproduced, and **two claims in the finding were measured wrong** — one in each direction.

**Wrong, and worse than stated:** the predicted outcome was that `container.Query("content")` would be empty. It is not — the count is 1. The element stays in the tree; what is disposed is the *fiber* that owns it. Effect cleanup runs, `PreviousTree` is nulled and its pooled nodes returned, and the element lives on outside the portal **owned by nothing, with its VNodes already back in the pool**.

**Wrong, and it changes whose defect this is:** the case does **not** fail on `origin/main`. Ported to a throwaway fixture on a clean main worktree it passes — neither `OwningPortalPlaceholder` nor `DisposeInlineFibersOwnedByPortal` exists there, so nothing disposes the re-homed fiber. **This branch introduced it.**

**The fix, and where the "remove a dependency" lesson actually applied.** Re-stamping on the carry path, because no non-stale key is available: containment cannot reach a *sibling* of what the teardown removes; `MountSlotStart` cannot, because `ShiftSlotStartsAfter` moves `PortalSlotInfo.SlotStart` without touching any fiber, and the trailing-empty-child case puts the fiber on the range's exclusive end where it is indistinguishable from the next portal's first child.

But re-stamping alone leaves the stamp load-bearing for **descendants**, and there it can never be repaired — a fiber mounted by a portal child's own isolated re-render sees `CurrentPortalPlaceholder == null`. Rather than copy the parent's stamp onto it (a second copy nothing keeps current), the disposal now walks `_parentToInlineFibers`, which is derived from the registry key and cannot go stale. **That index was written and maintained but never read**, while its comment claimed it existed "so subtree disposal locates them without scanning `_inlineInstances`". It now does what it said.

Each half cut separately reddens exactly its own case. The arrangement sweep found both arrangements load-bearing: skipping phase 2 is the whole discriminator for the first, and the portal child returning a **Fragment** rather than a wrapper `Div` is for the second — swapping in a `Div` makes it pass with the fix cut, because the containment sweep then reaches through the wrapper.

**A universals sweep caught one of my own sentences.** I had written that an isolated re-render "cannot observe a Portal the enclosing walk has since moved its parent out of". False: `ExpandFiberPreviousTree` descends into `PreviousTree` unconditionally, bail or not, so an inherited stamp *would* be repaired. The real reason — no second copy to keep current — is what is written now.

EditMode 3873/3873, PlayMode 161/161, no inconclusive.

s4k10503:
Checked whether #610 (a fiber carried to another container keeps `MountPoint` pointing at the one it was created in) belongs in this branch rather than as an issue. **It does not — it reproduces on a clean `origin/main` worktree at `6a3e461`, with none of this branch present**, byte-for-byte with the numbers in the issue:

```
PROBE rehome: container content=1 target content=0
PROBE swap:   container swapped=0 container content=1
              target    swapped=1 target    content=0
```

The whole mechanism is on `main` untouched by this branch: `LookupInline` keying on `(parentFiber, positionKey, identity)`, and `ReconcileExistingFiber` writing `existingFiber.MountSlotStart = site.SlotStart` while dropping `site.MountPoint`. This branch only adds the `OwningPortalPlaceholder` write beside that line. The probe never re-registers a portal id, so it exercises none of the retarget feature.

Evidence and the reusable probe fixture are on #610. Nothing to do here; no change made to this branch.

s4k10503:
## What the base-red check found, and what it was right about

`Base-red check (EditMode)` was not misfiring. Run locally against the merge base (`533d122`, `0` `error CS` in the base run's editor log), it named four cases of `PortalRegistryRetargetTests` that passed on a tree without any of this branch's production code:

```
passed on the base   Given_ASecondPortalOnTheReplacedElement_When_TheFirstFollowsItsId_Then_TheBridgeStays
passed on the base   Given_AComponentReHomedOutOfAPortalsChildren_When_ThatPortalCloses_Then_TheCloseTakesNothingFurther
passed on the base   Given_AComponentReHomedInThePortalsClosingRender_When_ThatRenderCommits_Then_ItSurvivesTheClose
passed on the base   Given_ANeighbouringPortalThatGrewFirst_When_ItCloses_Then_TheOtherPortalsComponentChildSurvives

4 case(s) the base already answers, or cannot be answered for.
```

All four had the same shape, and it is a shape four rounds of review did not catch: each asserts a **negative** — the bridge was not released, the re-homed component was not disposed, the modal's child was not disposed — and a base with none of the mechanism does not release or dispose anything, so the assertion holds for the one reason the case exists to rule out. The arrangement had no term in the assertion depending on it.

All four are **sharpened**, not declared. No `GREEN_ON_BASE` was added.

| case | what was folded in | on the base now |
|---|---|---|
| `…Then_TheBridgeStays` | the move itself, `Names(replacement)` | `Expected: (gone, True)` / `But was: (, True)` |
| `…Then_TheCloseTakesNothingFurther` | a portal child that *stays*, so the close has a disposal of its own | `Expected: (True, 0, 1)` / `But was: (False, 0, 1)` |
| `…Then_ItSurvivesTheClose` | the same staying child | `Expected: (True, 0, 1)` / `But was: (False, 0, 1)` |
| `…Then_TheOtherPortalsComponentChildSurvivesUntilItsOwnClose` | the modal's own close, after the toast's | `Expected: (0, 1)` / `But was: (0, 0)` |

Every case in the fixture is now red on the base, and `base_red_check.py --lane csharp --platform EditMode --base origin/main` exits `0` locally.

### Each sharpening is earned by a mutation, not by its shape

A case red on the base can still be insensitive to the code it sits next to, so each was mutated against this branch's own tree. Each mutation reddens exactly one case:

- **remove the carry-write** `existingFiber.OwningPortalPlaceholder = _ctx.CurrentPortalPlaceholder` in `ComponentRegistry.ReconcileExistingFiber` → only `…Then_ItSurvivesTheClose` dies, `(True, 0, 1)` → `(True, 1, 1)`. The leaver term is live and the phase-2 sibling is correctly not a stand-in for it.
- **widen the selection** in `DisposeInlineFibersOwnedByPortal` from `ReferenceEquals(fiber.OwningPortalPlaceholder, placeholder)` to "any portal-owned fiber" → only `…SurvivesUntilItsOwnClose` dies, `(0, 1)` → `(1, 1)`. It is not redundant with `…ThatClosedFirst…`, which survives this mutant; the two-portal pair now separates.
- **drop the survivor scan** in `FiberElementCleaner.ReleaseBridgeIfLastPortalOn` → only `…Then_TheBridgeStays` dies, `(gone, True)` → `(gone, False)`, with the move term still satisfied.

`…Then_TheCloseTakesNothingFurther` is the one I could not earn this way: no mutation I tried reddens it while leaving the others alone. It is red on the base, which is what the guard asks, but its discriminating power over this branch's code is unproven and I am saying so rather than implying otherwise.

### One reading is deliberately an occurrence, not a count

The staying child's disposal is asserted as `s_stayingCleanups > 0`. That is not laziness. Measuring it produced a finding, below.

### Suites

EditMode `3873/3873`, PlayMode `161/161`, `0` failed, `0` inconclusive, `0` skipped, `0` `error CS` in either log. `assert_no_inconclusive.py` clean on both.

No `CHANGELOG` entry: this change is test-side only and alters no behaviour a user can observe. No `Documentation~` change for the same reason.

---

### Found while measuring, and out of scope for this branch

A `V.Portal`'s top-level `V.Component` child is **re-created on every patch of the portal's children**, and the instances it leaves behind are never cleaned up. Measured on `origin/main` at `533d122` and on this branch, identically, by driving one portal through `n` patches of its children and then dropping it, counting mount-effect bodies and cleanups for a child whose deps are `Array.Empty<object>()`:

| patches before the close | mount-effect runs | cleanups on `main` | cleanups on this branch |
|---|---|---|---|
| 0 | 1 | 0 | 1 |
| 1 | 3 | 0 | 2 |
| 2 | 5 | 0 | 2 |

An effect with empty deps running again, with its previous cleanup never having run, means separate hook-state instances rather than one fiber re-running — two more of them per patch. So a component written directly under a portal does not keep its state across a patch of that portal's children, and the abandoned instances accumulate. The mount counts are the same on both sides, so this branch neither causes it nor is blocked by it; the branch's disposal reaches two of them at the close rather than all, which is why the case above reads an occurrence instead of a count. Worth its own issue.

Issue #596, opened 2026-08-10
BundledStyleSheetInclusionTests fails 9/11 in a git worktree and passes in the primary checkout


s4k10503:
`BundledStyleSheetInclusionTests` passes 11/11 in the primary checkout and fails 9 of 11 in a `git worktree` of the same commit. Measured on `origin/main` at `8bfbad8` with no local modifications, and reproduced independently by two separate runs — on a copied `Library`, on the same `Library` warm, and on a cold import.

```
BuildFailedException: Packages/com.velvet.core/Runtime/Assets/VelvetRuntimeAssets.asset did not load
```
preceded in the log by
```
No script asset for VelvetRuntimeAssets. Check that the definition is in a file of the same
name and that it compiles properly.
```

`AssetDatabase.LoadAssetAtPath<VelvetRuntimeAssets>` returns null in a worktree. The `.asset`'s `m_Script` GUID matches `VelvetRuntimeAssets.cs.meta`, so it is not a stale reference in git.

**This is not a red `main`.** CI checks out normally and its EditMode job is green; the primary checkout here is 11/11. What is broken is the checkout shape, and that is the one this repository uses for all branch work — `git checkout -b` is refused by a hook, so every branch is developed in a worktree. So every contributor running the suite the documented way sees nine failures and has to decide unaided that they are not theirs. Twice today an agent reported them as "pre-existing and not mine" only after building a pristine comparison tree to prove it, which is a cost paid on every run.

## What would close it

First establish why the load fails there — a worktree's `.git` is a file rather than a directory, and Unity's asset database, the `Library` it was cloned from, and any absolute path baked into an import can each react to that. Whether the fix is in the fixture, in how the repo prepares a worktree, or in `CONTRIBUTING.md` documenting a preparation step, depends on which of those it is.

Whatever the answer, the guard that makes it stick is that the failure must not be silent about its cause. A fixture that cannot load its own asset should say "this checkout cannot load it" rather than reporting a build failure that reads like a code defect — the whole cost here is nine reds that look like the contributor's fault.

Related: CLAUDE.md documents the headless run, and `velvet-open-followups` records the Library-precopy technique that makes a worktree run practical; neither mentions this.

s4k10503:
Still reproduces on current `main`, and the variable is the `Library`, not the commit.

[...]

PR #597, opened 2026-08-10
tooling(velvet): run a branch's changed test cases against the merge base

s4k10503:
`CLAUDE.md` asks that a regression test be verified red without the fix and green with it. Nothing
checked, so it was a matter of attentiveness. Run over four branches from recent history, three had a
case that passed on the merge base and had been reported as pinning something.

`scripts/test_quality/base_red_check.py` checks out the merge base, copies the branch's changed test
files onto it, and runs the cases the branch wrote there. A case green on both sides fails the check.

## 1. Where it runs

Measured on one machine, EditMode, warm `Library`:

| | |
|---|---|
| Full suite, 3844 cases | 84 s / 97 s |
| One fixture (27 cases), filtered | 15 s |
| A whole base check — build the tree, seed a `Library`, run the fixtures, decide | 26–47 s |
| A base tree with no `Library` at all — first import plus one fixture | 65 s |

So restricting to the changed fixtures is worth roughly 5× against the suite, and the base tree's
first import is the larger half of what remains. That settles the local shape: a base check costs
about a third of a suite run, which is cheap enough to run per branch.

CI is a different question, because the marginal cost there is a second Unity job — image, licence,
`Library` restore — not the test time. So the two lanes are split:

- **`Test ▸ base-red-python`** needs no licence and no editor, runs on every pull request and merge
  group, and is complete for the Python surface. It is in `required-checks`.
- **`Test ▸ base-red`** is a matrix job over EditMode/PlayMode, gated on `UNITY_LICENSE` and on
  `unity-tests`, and reaches Unity through the same `game-ci` action the suite does. It skips
  outright when the branch changed no C# test file. It is in `required-checks` too, because
  `RequiredCheckAggregationTests` refuses a job the aggregate does not depend on.

`--emit` / `--verdict` split the run so the editor can be somebody else's: the first builds the base
tree and writes what it read, the second decides from a results file. That split was exercised
locally end to end against the same commit as the single-shot run, and reached the same verdict.

## 2. Compilation

A new test on the base does not compile when it names a symbol the branch adds — the common case,
not the exception. **The verdict never reads the editor log**, which is what makes the reading
portable to a runner that hands back only a results file:

- a fixture the results name **no case of** is one the base did not build what was carried for. That
  is the evidence a test naming a new symbol is supposed to leave, so it is not a failure;
- a case missing while its fixture reported **its other cases** is a name that did not resolve, and a
  name nobody answered to is a reading nobody took. That fails.

What separates that from a genuine break is that the same file compiles on the branch, which
`needs: unity-tests` establishes; the reading is stated in the job's comment rather than assumed.

Two limits, stated rather than papered over:

- It does not by itself separate "the base could not build the branch's test file" from "the base
  could not build for a reason no branch owns". What closes that is the instrument reading: the cases
  the branch did **not** change come from the base's own green run, so one of those going red
  withdraws its fixture's verdicts. Where the branch adds a wholly new file and leaves no such case,
  three of the base tree's own fixtures are run alongside and at least one must pass.
- **A compile error stops the whole run, not one assembly.** Measured: breaking one test file and
  filtering to a fixture in a different assembly gives `Aborting batchmode due to failure: Scripts
  have compiler errors` and no results file at all. So a single round cannot say *which* carried file
  the base could not build — it can only say that it could not build one of them, which passes. The
  local harness withdraws one carried file per round and asks again until something runs, which is
  what turns that into a per-file verdict; a workflow does not express the loop, so CI runs one round.
  What survives in CI is the case the base **did** build and answer green — which is the case this
  exists for, and the one all three reproductions below turn on. What CI gives up is a green-on-base
  case sitting in the same assembly as an uncompilable sibling.

## 3. Legitimate green-on-base cases

A characterization test and a test riding with a pure refactor both belong and both pass on the base.
They say so in the file, directly above the case, per case:

```csharp
// GREEN_ON_BASE(characterization): the keyed-reorder order this refactor must not change.
```

The category must be one the script defines and the reason must be more than a placeholder; both are
checked. It is not a blanket opt-out in three ways: it covers one case and not its neighbours (a blank
line or any code between it and the attribute block ends it); it is read **in both directions**, so a
declaration over a case that turns out red on the base fails the check and has to be removed; and it
is prose a reviewer reads in the diff next to the case it excuses.

`BaseRedDeclarationTests` holds the categories written in C# sources, and the example this guide
shows, against the set the script actually accepts. `scripts/test_quality/test_base_red_check.py`
holds every declaration in the repository against the same reader, and fails on one that sits where
no case picks it up.

## 4. What "changed test" means

Covered, from `git diff` against the merge base (working tree included, so it runs before a commit):

- a case whose own lines the branch touched — a renamed method included, since the signature line is
  one of them, and a changed assertion likewise;
- every case of a file the branch adds whole.

Not covered, and **printed on every run** rather than left to be assumed:

- lines that fall inside no case — a `SetUp`, a field, a private helper the cases call, a `using`.
  Selecting the file whole on that ground was tried first: it puts all 27 cases of a fixture on trial
  for one line added to `SetUp`, and it consumes the untouched cases that the instrument reading needs.
  The run reports the count per file instead;
- a changed helper under `TestUtilities/`. It is carried onto the base so the cases compile, but the
  cases that call it are not selected;
- a production change with no test change at all. `mutation_check.py` is the instrument for that
  question, and this does not replace it — of the four branches below, the two whose reported defect
  was a surviving mutation rather than a green base are exactly the two this is weakest on.

## Proof against this repository's history

Each row is a real commit, `--base` its merge base, one C# base run:

| Branch state | Verdict | What it named |
|---|---|---|
| `#582` before the fix | **fails** | `Given_ConcurrentMutations_…_FinalStateComesFromLatestCall` **passed on the base** — its `mutationFn` returned a bare completion-source task, so the token it claimed to test was ignored either way. The sibling case added in the same commit came back red, so this is a reading within the commit, not a blanket one |
| `#582` after the fix | that case flips to **red on the base** | the run still fails on a fourth case in the same commit, `Given_AMutationInFlight_…_TheUnmountCompletes`, which is green on the base. The next commit on that branch rewrote exactly that case |
| `#582` after the fix, with a declaration on that case | **passes** | and with a deliberately wrong declaration over a red case, fails, naming it stale |
| `#583` | **fails** | `Given_ACheckoutAndAWorktreeOfIt_…_ItRefuses` passed on the base, along with four more in the same fixture. The guard the cases pose is absent from the base, so `python3 <missing file>` exits 2 and reads as a refusal |
| `#588` | **fails** | three of seven changed cases pass on the base |
| `#577` | **passes** | correctly: all five changed cases are red on the base. Its reported defect was mutation survival, which this does not measure |
| `#589` | **passes** | 29 cases across three fixtures, all "could not compile there"; with a fourth, compilable fixture added, the withdrawal loop converges in four rounds and measures it |

Two defects in this were found by running it, not by reading it, and both were the same shape — a
reading that comes back empty and passes:

- the fixtures it stands the base tree up against were being taken from `Library/PackageCache`, which
  compile into Unity's own assemblies and answer to no filter this project can pose, so the canary
  found nothing passing and withdrew the branch it was there to vouch for. Found by running it on
  itself;
- a case written in an abstract fixture (`PoolHelperTestsBase`, six cases, five heirs) is reported by
  the runner under each deriving class and never under the one it is written in, so the name taken
  from the file matched nothing and the case read as one the base could not build — which passes.
  `as_the_runner_names_them` resolves it, and a repository-wide case fails on any name no concrete
  class in the tree answers to.

`CONTRIBUTING.md` gains a section next to the mutation and neuter harnesses, and its CI table gains
rows for both new jobs — plus the `Test ▸ test-quality` row the table never had. No `Documentation~`
guide describes contributor tooling, so none changes. No `CHANGELOG.md` entry: nothing here is
observable to a package consumer.

## Known failures this does not introduce

The EditMode suite reports nine failures in `BundledStyleSheetInclusionTests` on this machine, all of
them `AssetDatabase.LoadAssetAtPath` returning null for
`Packages/com.velvet.core/Runtime/Assets/VelvetRuntimeAssets.asset`. They reproduce identically on
untouched `origin/main`, in a worktree with a copied `Library` and in one that imported from nothing,
and CI's own EditMode job is green — so they are local to this checkout shape and predate this branch.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -202,9 +202,108 @@ jobs:
             ${BASE:+--base "$BASE"}
 
   # ---------------------------------------------------------------------------
+  # A test written to pin a change has to be red without it, and nothing here asked. The Python half
+  # of that question is answered outright: base_red_check.py checks out the merge base, copies this
+  # branch's changed test modules onto it, and runs the changed cases there, which for Python is the
+  # whole run and costs seconds. The C# half needs an editor and is in `base-red` below.
+  #
+  # fetch-depth 0 because the merge base is what the question is asked against, and a shallow clone
+  # has no merge base to find.
+  # ---------------------------------------------------------------------------
+  base-red-python:
+    name: Base-red check (Python)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+
+      - env:
+          BASE: ${{ github.event.pull_request.base.sha || github.event.merge_group.base_sha }}
+        run: |
+          python3 scripts/test_quality/base_red_check.py --lane python \
+            --base "${BASE:-HEAD^}"
+
+  # ---------------------------------------------------------------------------
+  # The C# half of the same question. It reaches Unity through the same action the suite does, so it
+  # is split in two: the first step builds the base tree and writes what it read, the action runs the
+  # fixtures over that tree, and the last step takes the verdict from the results file. Nothing in
+  # between has to hand back an editor log.
+  #
+  # needs: unity-tests is load-bearing rather than tidiness. "It did not compile on the base" is only
+  # evidence about the branch where the same file compiles ON the branch, and that job is what says so.
+  #
+  # This is one round of a loop the local harness runs to a fixed point, and the difference decides
+  # what it can catch. A base that cannot build one carried file writes no results for anything, so a
+  # single round cannot tell which file that was: it reads as the base building none of the branch's
+  # tests, which passes. The harness withdraws one carried file and asks again until something runs,
+  # which is how it separates them — so `base_red_check.py` run locally is the sharper reading of the
+  # same branch, and what stays here is the case the base DID build and answer green, which is the
+  # one this exists for.
+  # ---------------------------------------------------------------------------
+  base-red:
+    name: Base-red check (${{ matrix.testMode }})
+    needs: [license-check, unity-tests]
+    if: needs.license-check.outputs.has_license == 'true' && github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        testMode: [EditMode, PlayMode]
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+          lfs: true
+
+      - id: read
+        env:
+          BASE: ${{ github.event.pull_request.base.sha }}
+        run: |
+          python3 scripts/test_quality/base_red_check.py --lane csharp \
+            --platform "${{ matrix.testMode }}" --base "$BASE" \
+            --base-tree base-tree --emit plan.json | tee reading.txt
+          grep '^fixtures=' reading.txt >> "$GITHUB_OUTPUT" || true
+
+      # After the reading, not before: the tree is a worktree git refuses to create over a directory
+      # that already holds a Library.
+      - uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        if: steps.read.outputs.fixtures != ''
+        with:
+          path: base-tree/Library
+          key: BaseLibrary-${{ matrix.testMode }}-${{ github.event.pull_request.base.sha }}
+          restore-keys: |
+            BaseLibrary-${{ matrix.testMode }}-
+
+      # continue-on-error because a base that cannot build the branch's test file is the answer, not
+      # a broken job. Which of the two it was is settled by the step after this from the results file.
+      - uses: game-ci/unity-test-runner@0ff419b913a3630032cbe0de48a0099b5a9f0ed9 # v4
+        if: steps.read.outputs.fixtures != ''
+        continue-on-error: true
+        env:
+          UNITY_LICENSE: ${{ secrets.UNITY_LICENSE }}
+          UNITY_EMAIL: ${{ secrets.UNITY_EMAIL }}
+          UNITY_PASSWORD: ${{ secrets.UNITY_PASSWORD }}
+        with:
+          # The version unity-tests runs, for the reason stated there.
+          unityVersion: 6000.3.21f1
+          testMode: ${{ matrix.testMode }}
+          projectPath: base-tree
+          artifactsPath: base-results
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          customParameters: -testFilter "${{ steps.read.outputs.fixtures }}"
+
+      - if: steps.read.outputs.fixtures != ''
+        run: |
+          python3 scripts/test_quality/base_red_check.py \
+            --verdict plan.json --results base-results
+
+  # ---------------------------------------------------------------------------
   # mutation_check.py needs a licence to kill a mutant, but not to build one, and the building is
   # the half everything else rests on: an operator that emits uncompilable C# reports that mutant as
-  # noise rather than as a survivor, so a real hole hides behind it. That half runs here.
+  # noise rather than as a survivor, so a real hole hides behind it. That half runs here, and
+  # base_red_check.py's reader is here for the same reason: a case its parser misses is a case the
+  # base is never asked about, and the run comes back green having measured one fewer thing.
   # ---------------------------------------------------------------------------
   test-quality:
     name: Test quality
@@ -215,6 +314,8 @@ jobs:
       - run: python3 scripts/test_quality/test_mutation_check.py
 
       - run: python3 scripts/test_quality/test_neuter_check.py
+
+      - run: python3 scripts/test_quality/test_base_red_check.py
 
   # ---------------------------------------------------------------------------
   # The one check branch protection requires from this workflow. Naming a real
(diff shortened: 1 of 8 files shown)


s4k10503:
Review round 1. Three HIGH and three MEDIUM, all confirmed and fixed — and the three HIGH turned out to be one condition, unstated.

**The success condition, written down before touching any input:** the run passes only where **every changed case was measured on a base tree that demonstrably answers.** `COULD_NOT_COMPILE` is indistinguishable, from a results file alone, from a base that built nothing, ran nothing, or was never asked. HIGH 1 is that condition unenforced at the instrument; HIGH 2 and HIGH 3 are it *unenforceable*, because the case was asked about under a name nothing answers to and no-reading falls through to the same passing verdict. Two doors, one failure mode.

Every fix is pinned by a case failing on a surgical revert of exactly that fix, all others in place. The two guards that could not fail are shown failing on the real corpus first: `homeless` reads the cases as written instead of the naming's output, and goes red at **475** and at **8** — the review's own numbers, reproduced.

**A defect underneath HIGH 2 the review did not name.** Fixing the type scan to read code lines *lost 13 cases*. `mutation_check.code_mask` treats the apostrophe in `#region Boundary's own normal mount` as opening a character literal and blanks to the next apostrophe anywhere in the file:

```
line 68  masked-out chars: 15 / 15  |  [Test]
line 69  masked-out chars: 92 / 92  |  public void Given_BoundaryComponent_When_MountedWithoutError_Then_ShowsNormalTree()
```

Twelve `ErrorBoundaryTests` cases had their `[Test]` invisible and thirteen more had truncated spans, so a branch editing line 75 of a 68–77 case was attributed to no case at all. **The mutation harness had this identically** — a masked offset generates no mutant, silently. Fixed there; 884 names corrected, 13 spans lengthened, case count unchanged at 3890.

**LOW 2 from the review is false and I changed nothing for it.** It reads the wrong column: the header is `| Workflow | Trigger | Unity license | Required to merge |`, and `base-red` genuinely needs a licence while `base-red-python` does not. Both say `no` under *Required to merge*, correctly.

**Left undone, deliberately, and it is in the unsafe direction.** The Python lane has no canary. `run_python` reports a non-zero exit as `Failed`, so "the base could not import this module" and "this case genuinely failed" are one verdict — measured on this branch, 37 of 38 changed cases came back red for one shared reason (`base_red_check.py` does not exist on the merge base), and a genuinely vacuous new Python case would have read the same and passed. Same family as HIGH 1. The remedy is the C# lane's, but adding a new mechanism unreviewed at the end of a round is how the last four guards shipped broken, so it is stated here rather than slipped in.

EditMode 3856 (9 pre-existing `BundledStyleSheetInclusionTests` failures, verified identical on a detached `origin/main` worktree — #596), PlayMode 161/161, `test_base_red_check.py` 70 OK (was 53), `test_mutation_check.py` 15 OK (was 12).

s4k10503:
A regression this branch introduced, and the guard that should have caught it.

**The reader dropped 15 cases across 6 files.** The new type-stack unwind is keyed on a body closing, so a type declared without a brace body is pushed and never popped — and because the pop loop stops at the first entry still open, one such declaration holds down every type under it too. This repository writes its store state as a positional record beside or among its fixtures (37 of them), so the cases standing under one were emitted under an owner chain nothing declares:

```
Velvet.Tests.struct.ButtonChildPoolReuseTests.Given_AButtonCarryingAChild_When_ResetForReuse_Then_ItHasNoChildren
Velvet.Tests.ComponentPropsComparerTests.SimpleProps.Given_SameReference_When_ShallowEquals_Then_IsEqual
```

`-testFilter` matches neither, so the fixture reports nothing, and a fixture that reported nothing is read as one the base could not build — `COULD_NOT_COMPILE`, which is not a failing verdict. The controls in the same file resolve to `None` as well, so `unsound_fixtures` clears them, and the canaries pass because they are other files. The run comes back green having measured none of those cases.

Worth stating plainly: `ButtonChildPoolReuseTests.cs` is the template CLAUDE.md names for pool-reuse regressions. A branch sharpening a case there passed this gate having measured nothing it changed.

**The repository-wide guard could not see it.** It read only the last owner segment, so `Velvet.Tests.struct.ButtonChildPoolReuseTests` passed on `ButtonChildPoolReuseTests` being a declared class. It now reads the whole chain against the qualified names the tree spells, and those are computed by matching each declaration's own braces rather than by a second running stack — a set of legitimate chains built from a stack that fails to unwind agrees with every chain such a stack emits and holds nothing. Widened first, it fails on the real corpus with exactly the 15, and passes the 26 cases that are genuinely nested.

**Also fixed:** `record` takes an optional `class` or `struct` before the name, so the declaration expression was reading the keyword as the name. Nothing here declares a bodied `record struct` today and the guard cannot hold it — it reads declarations through the same expression and would agree — so it is pinned by a case instead.

The reader finds 3915 cases before and after, losing and gaining none, with every span unchanged.

Python suites: `test_base_red_check` 73 (3 new), `test_mutation_check` 15, `test_neuter_check` 8, `test_duplication_check` 6, `test_unreadable_state_check` 12, `test_settle` 18, `test_published_check` 17, `test_release_notes` 26 — all green. No CHANGELOG entry: contributor tooling. No `Documentation~` guide describes this check; CONTRIBUTING.md's one-line summary of the guard is still accurate.

Not fixed, and worth knowing: `concrete_heirs` qualifies a heir as `namespace.SimpleName`, which is wrong for a nested heir such as `VariantElementLocalEdgeTests.BoundsGated`. No abstract fixture holding cases has a nested heir today, so nothing is misnamed — and the widened guard now fails if one appears, rather than passing in silence. Separately, both the old and the new reading of `abstract` look only at the declaration's own line, so a modifier split across lines would read as concrete.

PR #607, opened 2026-08-10
fix(velvet): key the suspended mark per Suspense, not per boundary

### LINKED ISSUE #604: A sibling Suspense that resolves clears the boundary flag its suspended sibling set
`SuspenseFallbackShown` is keyed `(boundary, positionKey)`; `SuspendedBoundaries` is keyed by **boundary alone**. The two are written as adjacent lines in `GeneralPathReconciler.cs:1043` and `:1048-1049`, and the second is unconditional per Suspense node:

```csharp
if (suspended) _ctx.SuspendedBoundaries.Add(boundaryFiber);
else _ctx.SuspendedBoundaries.Remove(boundaryFiber);
```

One component fiber rendering **two** Suspense nodes gets two distinct `stateKey`s (the node index is in the key) but one boundary entry. If A suspends and B does not, and B is expanded after A in the same new-side walk, **B's `else` branch clears the boundary while A's fallback is committed.**

The reader is `FiberWorkLoop.FlushState:211-227`. With the boundary cleared it stops deferring A's offscreen primary descendants, so a dirty descendant flushes into the slot range A's fallback occupies.

The intended invariant is `SuspendedBoundaries = { b : ∃k. SuspenseFallbackShown[(b,k)] }` — a derived reading, maintained by hand and derivable only from the finer-keyed table. There is a **third** representation of the same fact, `ComponentFiber.IsOffscreen` (`GeneralPathReconciler.cs:1021-1024`).

Verified by reading both the writer and the reader. **No test was written** — that is the first work here.

## What would close it

A per-boundary record holding a `Dictionary<string, bool>` with a derived `AnyFallback` makes the disagreement unrepresentable, and removes the third name with it. That is cheaper at runtime than today's two lookups, not more expensive.

Related: #603 — this is the clearest instance of the pattern that issue is about, and the only one measured to be a correctness defect rather than a leak.

s4k10503:
Closes #604

## What was wrong

`ExpandSuspenseInline` recorded each `V.Suspense`'s children-vs-fallback decision under `(boundary fiber, scoped position key)`, but the mark `FiberWorkLoop.FlushState` reads to defer an offscreen descendant's lane flush was keyed by boundary alone and written unconditionally per Suspense node:

```csharp
if (suspended) _ctx.SuspendedBoundaries.Add(boundaryFiber);
else _ctx.SuspendedBoundaries.Remove(boundaryFiber);
```

One component fiber can render several Suspense nodes, and they share that boundary fiber. With a suspended node expanded before a resolved one, the resolved one's `else` cleared the mark while the first one's fallback was committed. `FlushState` then stopped deferring the first one's offscreen primary descendants, so a state update inside those hidden children committed into the slot range the fallback occupied and the committed fallback left the tree.

Ordering decided it: the same two boundaries with the resolved one placed first were measured unaffected.

## What changed

Boundary state is now one table — per boundary fiber, the position keys of the Suspense nodes currently showing their fallback. Only fallback-showing keys are stored and a boundary with none holds no entry, so both readings `FlushState` needs ("this boundary has a fallback up" and "any boundary does") are derived from it rather than tracked beside it. A sibling Suspense expanded later writes its own key and cannot clear another's. Reads that previously did two dictionary lookups now do one, and pruning a disposed boundary is a single `Remove` instead of a scan of every key.

`ComponentFiber.IsOffscreen` is kept and is not a third copy of the same fact. It is per-fiber, and it is what separates a suspended Suspense's primary subtree from that Suspense's visible fallback subtree and from a sibling Suspense's children — all of which sit under the same boundary fiber, which the boundary-keyed reading cannot distinguish. Dropping the `IsOffscreen` term from the guard turns two cases red, so it is load-bearing in both directions.

## Tests

`SuspenseBoundaryTests` gains two cases over one host rendering a suspending `V.Suspense` followed by a resolving one:

- an offscreen primary descendant's state update leaves the committed fallback in place — red before this change (`Expected: not null / But was: null` on the fallback label);
- a descendant of the resolved sibling still flushes, so keeping the first one's mark up does not over-defer.

EditMode 3854 passed / 0 failed / 0 inconclusive; PlayMode 161 / 0 / 0.

## Documentation

No guide change. No `Documentation~` page describes the flush-deferral guard, and `V.Suspense`'s own remark about sibling boundaries being tracked independently was already accurate for fallback selection — the defect was in the flush guard. CHANGELOG has an entry under `[Unreleased]` / Fixed: the lost fallback is user-visible.

### DIFF
--- a/Packages/com.velvet.core/CHANGELOG.md
+++ b/Packages/com.velvet.core/CHANGELOG.md
@@ -38,6 +38,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Two `V.Suspense` boundaries in one component's render no longer share a single suspended mark. One
+  showing its fallback while a second, placed after it, rendered its children left the component
+  unmarked, so a state update inside the first one's hidden children stopped being deferred and
+  committed into the slot range its fallback occupied — the fallback disappeared from the tree.
+  Whether a boundary is showing a fallback is now read off the per-boundary record each `V.Suspense`
+  writes under its own position, so a sibling that resolves cannot clear it. Ordering decided it:
+  the same two boundaries with the resolved one placed first were unaffected.
+
 - A second `Mutate` call no longer cancels the first or drops its callbacks. Both run, each delivers
   its own `OnSuccess` / `OnError`, and `Status` / `Data` / `Variables` show the most recent — which
   is what TanStack Query does, and it hands `mutationFn` no signal at all. A double-tapped Buy used to
--- a/Packages/com.velvet.core/Runtime/Component/ComponentFiber.cs
+++ b/Packages/com.velvet.core/Runtime/Component/ComponentFiber.cs
@@ -185,12 +185,13 @@ namespace Velvet
         public bool IsSuspenseBoundary { get; internal set; }
 
         /// <summary>
-        /// True while this fiber is a primary (hidden) child of a wrapper-less Suspense boundary that
-        /// is currently showing its fallback. Set by <c>GeneralPathReconciler.ExpandSuspenseInline</c> when
-        /// the boundary suspends and cleared when it reveals. <see cref="FiberWorkLoop.FlushState"/>'s
+        /// True while this fiber is a primary (hidden) child of a wrapper-less Suspense that is currently
+        /// showing its fallback. Set by <c>GeneralPathReconciler.ExpandSuspenseInline</c> when that
+        /// Suspense suspends and cleared when it reveals. <see cref="FiberWorkLoop.FlushState"/>'s
         /// offscreen guard defers a lane flush for offscreen fibers (their slot is occupied by the
-        /// fallback) while still allowing the visible fallback subtree to flush (the
-        /// fallback renders normally; only the primary subtree is offscreen).
+        /// fallback). It is per-fiber rather than per-boundary because one component fiber can render
+        /// several Suspense nodes: that Suspense's own visible fallback subtree, and a sibling Suspense's
+        /// children, sit under the same boundary and must still flush.
         /// </summary>
         internal bool IsOffscreen { get; set; }
 
--- a/Packages/com.velvet.core/Runtime/Component/ComponentFiber.cs
+++ b/Packages/com.velvet.core/Runtime/Component/ComponentFiber.cs
@@ -186,8 +186,9 @@ namespace Velvet
 
         /// <summary>
         /// True while this fiber is a primary (hidden) child of a wrapper-less Suspense that is currently
-        /// showing its fallback. Set by <c>GeneralPathReconciler.ExpandSuspenseInline</c> when that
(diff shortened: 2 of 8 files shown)


s4k10503:
The `Base-red check (EditMode)` refusal at `4614343` is **correct**, not an unmeasured verdict — and `Required checks (Unity)` failed only as its aggregate, not from a second cause (`RESULTS: success success success success success success failure`, one entry).

The base tree demonstrably answered. Reproduced locally with the same script:

```
EditMode attempt 1: 45 case(s) over 4 fixture(s) in 28s
red on the base     ..._When_OffscreenDescendantUpdates_Then_FallbackSurvives  (failed)
passed on the base  ..._When_ResolvedSiblingsDescendantUpdates_Then_ItRerenders
```

`testcasecount="45" passed="44" failed="1" inconclusive="0"`, and `grep -c "error CS"` on the base editor log is **0**. The compile-failure reading does not apply: only test-side files are carried onto the base, and `SetSuspenseFallbackShown` is production code in `ReconcilerContext.cs`, so it is never carried and nothing referenced it there.

The red case fails on the base with:

```
An offscreen primary descendant's flush is deferred, so the committed fallback stays in its slot
  Expected: not null
  But was:  null
```

**The green case cannot be sharpened into a red**, because the base gets that outcome right — by never deferring at all, since the boundary-wide mark this branch replaces had already been cleared by the resolved sibling's own expansion. So it is a characterization case, and it is now declared as one.

Before declaring it I measured that it carries weight, with two mutations of the branch tree (full `SuspenseBoundaryTests` run each time, 0 compile errors):

| mutation | `..._FallbackSurvives` | `..._ItRerenders` | other cases |
|---|---|---|---|
| `FiberWorkLoop`: drop the per-fiber `IsOffscreen` walk, defer boundary-wide | Passed | **Failed** | 1 other |
| `GeneralPathReconciler`: mark `IsOffscreen` from the boundary rather than from this Suspense's `suspended` | Passed | **Failed** | **none** |

The second mutation reddens this case and **no other case in the fixture**, so it is not redundant with `Given_StatefulFallback_When_FallbackStateUpdatesWhileSuspended_Then_FallbackRerenders` (which dies under the first mutation but survives the second). Its failure text both times:

```
A descendant of the resolved sibling Suspense is not offscreen, so its own flush still commits
  Expected: not null
  But was:  null
```

`f15bff1` adds the declaration. `base_red_check.py` now exits 0: `red on the base` for one case, `declared, and green as declared` for the other.

Full suites on `f15bff1`, headless, no `-nographics`, 0 `error CS` in both logs:

- EditMode 3858/3858 passed, 0 failed, 0 inconclusive
- PlayMode 161/161 passed, 0 failed, 0 inconclusive
- `assert_no_inconclusive.py` clean on both

No CHANGELOG entry: the change is a test-side annotation with no user-visible behaviour. No doc update: CONTRIBUTING.md already owns `GREEN_ON_BASE` and this change only uses it.

Issue #601, opened 2026-08-10
A component re-homed out of a portal loses its state and effects, and is unowned for one render


s4k10503:
Move a **keyed** `V.Component` out of a portal's children into a sibling position under the same declaring component, and its fiber does not survive — although the registry key is identical on both sides.

Measured (`portalCParent == hostFiber == movedCParent`, same explicit key, same identity):

```
P2  setups=1  cleanups=1  reg=[host only]     <- effect cleanup ran at the move; element mounted, no registered fiber
P3  setups=2                                   <- a fresh fiber, setup runs again
```

So state and effects die silently on a re-home, and there is a **one-render window where the element is live and owned by nothing**.

`ComponentRegistry`'s key is `(parentFiber, positionKey, identity)` and `positionKey` is `component.Key` when explicit, so the lookup should hit. It does not, and the cause is not the key.

Reproduced with PR #593's production change reverted, so this is `main`'s behaviour, not that branch's.

## Why it matters beyond the leak

`ComponentRegistry.cs:266-269` currently gives as the sole reason a write-once portal stamp is safe that "a component that moved out of a Portal's children misses the lookup and registers as a new fiber". The conclusion is right — the fiber does not survive — but the stated mechanism is wrong, and PR #593 is correcting that comment rather than the behaviour. Anything else that reasons from "the key differs on a re-home" is reasoning from a false premise.

## What would close it

Establish why `LookupInline` misses when the key is equal, then decide whether a re-home should preserve the fiber at all. React's answer for a keyed element moving between parents is that it does not — the subtree unmounts and remounts — so preserving may not be the goal; if it is not, the defects are the silent part and the unowned render, not the loss.

s4k10503:
Two independent measurements of this, from separate sessions. They agree on the outcome and differ on one clause, and the difference matters for anyone acting on this issue.

**Agreed, by both:** a keyed component re-homed out of a portal's children does **not** keep its fiber. The old one is disposed and a new one registers.

- reading A: `P2 setups=1 cleanups=1 reg=[host only]` — effect cleanup at the move, then a fresh fiber
- reading B: `cleanups after1=0 after2=1 after3=1, stamped1=1 stamped2=0` — the move disposes the stamped fiber and registers an unstamped one

[...]

Issue #409, opened 2026-08-06
Unkeyed component instance identity ignores which container the component renders into


s4k10503:
> ### Re-measured against `origin/main` (`89042a0`) on 2026-08-13 — the direction is now inverted
>
> **The live container is now the RIGHT one, not the left.** PR #612 added
> `if (site.MountPoint != null) existingFiber.MountPoint = site.MountPoint;`
> (`ComponentRegistry.cs:194`), so the shared fiber's `MountPoint` follows whichever container
> reconciles it **last**. A `setState` from either occurrence now re-renders into the *last*
> container and the *first* goes stale — the mirror image of what the empirical section below
> recorded.
>
> Measured by #612's own author across the same shape:
>
> | | before #612 | `origin/main` |
> |---|---|---|
> | mount | `left=[n0] right=[n0]` | `left=[n0] right=[n0]` |
> | after `setState(1)` | `left=[n1] right=[n0]` | `left=[n0] right=[n1]` |
> | after `setState(2)` | `left=[n2] right=[n0]` | `left=[n0] right=[n2]` |
>
> Both are wrong. What changed is only which container goes stale: `MountSlotStart` already named
> the last container reconciled, and `MountPoint` now agrees with it — the two position fields no
> longer contradict each other, which is a precondition for fixing this rather than a fix of it.
>
> The defect itself is untouched: the registry key still omits the mount point, and the two
> occurrences still share one fiber and one set of hooks. The workaround is unchanged — give each
> occurrence a distinct `key:`.
>
> The empirical readings further down are kept as the record of what was observed when this was
> filed; read every `a=…`/`b=…` in them with the containers swapped.

## What React does

React ties a component instance (and all its hook state) to its position in the rendered tree. `<div><Counter/></div><div><Counter/></div>` produces two independent Counter fibers with independent state — the child fiber list of each `div` is its own reconciliation scope (react.dev, "Preserving and Resetting State": "React keeps the state around for as long as you render the same component at the same position in the tree").

## What Velvet does

[...]

PR #612, opened 2026-08-11
fix(velvet): key a portal's component children on the fiber that declared it, and carry the mount point

### LINKED ISSUE #610: A fiber carried to another container keeps MountPoint pointing at the one it was created in
`ComponentFiber.MountPoint` is written only at creation, by `FiberRenderer.MountInline` / `Mount`. `ReconcileExistingFiber` updates `MountSlotStart` and drops `site.MountPoint`. So a fiber carried to a **different container** — same explicit key, same parent fiber, same identity, so the registry lookup hits — keeps pointing at the container it was created in, and its own isolated re-render reconciles into that one.

Measured with a probe (a component re-homed out of a portal, then re-rendering itself with a different output):

```
after rehome: container content=1  target content=0        <- the carry itself is correct
after swap:   container swapped=0  container content=1
              target    swapped=1  target    content=0     <- written into the wrong container
```

The component re-homed into `container`, re-rendered itself, and wrote its new output into the **portal target**, leaving the stale `content` element behind in `container`.

**This is wider than portals.** Any inline fiber that keeps its registry key while moving between two host containers hits it; the portal case is just the one that made it visible. It also means a stale `MountPoint` can get a re-homed fiber disposed by `DisposeFibersUnder` containment if the old container is torn down later.

## Why it was not fixed where it was found

`MountPoint` is read by `FiberEffects`, `FiberBatchScheduler`, `FiberCommitWork`, `FiberCrossPanelInput`, `FiberContextSpine` and the containment sweep. **Whether each of those six wants the current position or the creation position is the question**, and it is not answerable from the portal change that surfaced this. Bundling a guess into a portal-stamp fix is how PR #593 spent four rounds.

## What would close it

Answer the six readers first, one at a time, with the evidence for each. Then decide whether `MountPoint` becomes current-position (and `ReconcileExistingFiber` maintains it), or whether the readers that want the current position stop reading `MountPoint` and ask something else.

Pin the probe above either way — it needs no portal beyond getting the carry to happen.

Related: #603 is the family (a position represented in several places, none authoritative); #601 is the neighbouring question of whether a re-homed fiber should survive at all.

### LINKED ISSUE #611: A portal's top-level component child is re-created on every patch, losing its hook state
A `V.Component` written as a top-level child of a `V.Portal` is **re-created on every patch of the portal's children**, so its hook state and refs do not survive a patch.

Measured — the component's mount effect runs:

| patches of the portal's children | effect runs |
|---|---|
| 0 | 1 |
| 1 | 3 |
| 2 | 5 |

Identical on `origin/main` and on the `feat/registry-portal-retarget` branch, so this is not from any open work.

Two runs per patch is the shape of a create-and-discard followed by a create: the fiber is not found, a new one is made, and the previous one's cleanup runs.

## Why it matters

`Documentation~/portals.md` says of a *move* between containers that "Moving is an unmount and a remount, so child state, refs and effects do not survive it" — presented as the price of a move. This is an ordinary **patch**, where nothing moved and the guide promises nothing. A `UseState` inside a portal's component child silently resets whenever anything else in that portal re-renders.

## What would close it

Find why the inline lookup misses across a portal-children patch. `PatchPortalChildren` pushes the declaring fiber (`FiberNodePatcher.cs:750`), so the `(parentFiber, positionKey, identity)` key ought to match on the second pass — establish which of the three differs, and whether it differs for the same reason as #601 (where a re-homed component's key is provably identical on both sides and the lookup misses anyway).

Those two are likely one defect. Worth checking before fixing either.

Reproduction by the agent working on PR #593; filed separately so that branch stays scoped.

Related: #601, #603.

s4k10503:
A `V.Component` written as a top-level child of a `V.Portal` did not survive a patch of that
portal's children. Its hook state reset, its effects re-ran, its previous instance's cleanups never
ran at all, and the element it had rendered stayed behind in the target.

`Documentation~/portals.md` opens by promising that a portal's children "stay part of the logical
tree — context, state and re-renders flow from the call site". That held for every child position
except this one.

## Why the lookup missed

`ComponentRegistry` keys an inline-mounted fiber on `(parentFiber, positionKey, identity)`. A
portal's top-level children mount in `ChildReconciler.DrainPendingPortalMounts`, which runs from the
top-level reconcile's `finally` — long after the declaring component's own fiber came off
`FiberStack`, so `FiberStack.Current` there is the reconcile root. Every later patch of the same
portal reaches `FiberNodePatcher.PatchPortalChildren` from the reconcile of the tree the declaring
component itself returned, with that component's own fiber current.

The registry's inline table across a mount and the first patch, on `main`:

```
after mount:     parent=F(<Mount>b__0)  pos=c  ident=DiagChild  -> F252644992(DiagChild)
after 1 patch:   parent=F(<Mount>b__0)  pos=c  ident=DiagChild  -> F252644992(DiagChild)
                 parent=F(HostKeyed)    pos=c  ident=DiagChild  -> F-2127519168(DiagChild)
```

Same `positionKey`, same `identity`, different `parentFiber`. The patch looks under the declaring
fiber, finds nothing, and builds a second instance. The mount-time entry is never reached again and
never disposed, so its element sits in the target for the life of the tree.

The drain now pushes the declaring fiber it already captures at enqueue (`LogicalParent`, which
`PortalState.DeclaringFiber` records anyway) around its nested reconcile, so both halves of the
portal's life agree on the registry parent. This is the same push, for the same registry-parent
reason, that `Reconciler.BeginDetachedItemScope` already does for a `V.VirtualList`'s
controller-mounted items.

## The measured cost, corrected

#611 reports the child's mount effect running 1 / 3 / 5 for 0 / 1 / 2 patches — two extra instances
per patch, without bound. That is not what happens. Measured on `main` across four host shapes — an
unkeyed child, a keyed child, a portal nested under a `V.Div`, and a portal whose only child is the
component — the count is **1 / 2 / 2 / 2**: one extra instance at the *first* patch, then stable,
because the second instance registers under the declaring fiber and every later patch finds it.

What is unbounded is not the instance count but the consequence: cleanups stay at 0 forever, and the
target keeps an element nothing owns. A portal declared as
`children: [V.Component(Child), V.Div(name: "sib-" + n)]` read `content|sib-1|sib-0` after one patch
instead of `content|sib-1`.

## #601 is a different defect, and half of it is not a defect

#611 asked whether it and #601 share a cause. They do not, and the two sentences are different:

- **#611** — the lookup is reached and misses on `parentFiber`, per the table above.
- **#601** — the key is identical on both sides, the lookup is reached, and it misses because the
  entry has *already been removed*. The departing container's own reconcile disposes the fiber as an
  orphan before the walk reaches the component's new position. An ordered event log across the
  re-home render reads `cleanup, body, setup`.

The control that separates them carries no portal at all: a keyed `V.Component` moved from
`V.Div(name: "A")` to a sibling `V.Div(name: "B")` under the same declaring component produces the
identical reading — `cleanup, body, setup`, one setup gained, one cleanup run, and a fresh fiber
under the byte-identical registry key. So a keyed component changing parent is unmounted and
remounted whether or not a portal is involved, which is what React does for a keyed element that
changes parent and what the portals guide already states of a move.

What #601 additionally observed — an element mounted in the target with no fiber registered for it —
is the leak this PR fixes, not a second mechanism. With the fix the portal target holds nothing of a
component that has re-homed out of it; before it, the target kept the mount-time instance's element.
That case is pinned here. The remaining loss of state on a re-home is left as it is.

## #610 is the other half of the same key, so it is fixed here

`ReconcileExistingFiber` is one method, and it writes three fields that all say where a fiber's
output lives. On a carry it updated `MountSlotStart` and `OwningPortalPlaceholder` and left
`MountPoint` at the container the fiber was created in — so a fiber carried elsewhere re-rendered
itself into the container it had left, at the *new* container's slot index. Splitting that into a
second pull request would mean shipping a registry whose position fields disagree with each other.

Every reader of `MountPoint` in `Runtime/` wants the current position:

| reader | what it does with the VE |
|---|---|
| `ComponentRegistry.DisposeFibersUnder`, `GeneralPathReconciler` orphan sweep | containment test for "is this fiber inside a subtree leaving the DOM" |
| `FiberRenderer.Unmount`, `FiberErrorBoundary` | reconciles the fiber's own committed output away / into a fallback |
| `FiberWorkLoop`, `FiberCommitWork` isolated re-render | `Reconcile(MountPoint, …, slotStart: MountSlotStart)` — paired with an already-current field |
| `FiberCommitWork.NextInlineSiblingSlotStart` | groups co-located siblings, compared against `MountSlotStart` |
| `FiberWorkLoop.IsHostDetachedFromRoot` | VE-root comparison deciding whether to suppress a flush |
| `FiberEffects`, `FiberBatchScheduler`, `Hooks.UseFrame`, the `schedule.Execute` sites | reaches `.panel` / `.schedule` |
| `FiberCrossPanelInput` | start element for walking outward to the logical chain |

The one that could want the creation position, `FiberContextSpine.PushOutletHost`, reads a
**wrapper-mounted** fiber's `MountPoint`. A wrapper-mounted fiber's registry key *is* its wrapper VE,
so a different container is a different key: it structurally cannot see a carried fiber.

## What this does not fix, and what it changes there

#409 — two components at the same unkeyed position in *different* containers sharing one instance —
is untouched and stays open. This is the same key failing from the other end, and #409 already names
this consequence: "the fiber's `MountPoint` stays the first container, so a `setState` in either
re-renders only the left panel".

Which container goes stale does change, and it was measured rather than reasoned:

| | `main` | here |
|---|---|---|
| mount | `left=[n0] right=[n0]` | `left=[n0] right=[n0]` |
| after `setState(1)` | `left=[n1] right=[n0]` | `left=[n0] right=[n1]` |
| after `setState(2)` | `left=[n2] right=[n0]` | `left=[n0] right=[n2]` |

Both are wrong while #409 stands. The difference is that `MountSlotStart` was *already* the right
panel's, so `main` reconciles into the left container at the right container's slot index; the two
position fields now agree, which is a precondition for fixing #409 rather than a fix of it. The
workaround is unchanged: give each occurrence a distinct `key:`.

## Tests

`PortalChildFiberContinuityTests` (EditMode). Five cases red on the base with production reverted and
the tests unchanged — the first four with all production reverted, the fifth with only
`ComponentRegistry` reverted, so it is measured against the portal fix rather than through it:

```
Given_..._Then_ItsMountEffectDoesNotRunAgain          Expected: (1, content|sib-1)  But was: (2, content|sib-1|sib-0)
Given_..._Then_TheTargetKeepsOneElementForIt          Expected: "content|sib-1"     But was: "content|sib-1|sib-0"
Given_..._Then_ItKeepsThatState                       Expected: (7, content|sib-1)  But was: (0, content|sib-1|sib-0)
Given_..._Then_TheTargetKeepsNoElementForIt           Expected: (0, 1)              But was: (1, 1)
Given_..._Then_ItWritesIntoTheContainerItIsIn         Expected: (b, c, )            But was: (b, b, c)
```

That last one reads the mark after the carry, the container the component now occupies, and the one
it left. The post-carry term is load-bearing: a fiber rebuilt instead of carried would reach `c` in
the right container too, and would read `a` there.

A sixth case pins that the fiber the drain pushes comes back off the stack. It carries a
`GREEN_ON_BASE(characterization)` declaration, because it pins a balance the base does not have —
the base pushes nothing there, so it is even for want of a push rather than by unwinding one.
`scripts/test_quality/mutation_check.py` measured that removing the matching `Pop` leaves the whole
EditMode suite green, which is why the case exists.

EditMode 3885 passed / 0 failed / 0 inconclusive; PlayMode 161 passed / 0 failed / 0 inconclusive.

## Documentation

No guide change. `Documentation~/portals.md` already states the contract this restores, and already
states that a move between containers is an unmount and a remount — the behaviour the control above
confirms is unchanged. `CHANGELOG.md` gains an entry under `[Unreleased] / Fixed`, since a component
child of a portal keeping its state across a patch is user-visible, as is a carried component
re-rendering into the container it occupies — including the note about which container goes stale in
the shared-instance case above.

Closes #611
Closes #610

### DIFF
--- a/Packages/com.velvet.core/CHANGELOG.md
+++ b/Packages/com.velvet.core/CHANGELOG.md
@@ -37,6 +37,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- A component the reconciler carries to a different container now re-renders itself into the container
+  it is in. Only its slot index followed the move, so its own `setState` reconciled its new output into
+  the container it had left — leaving the stale element behind there and writing the new one into a
+  container it no longer occupies. It also let a component that had moved out of a subtree be disposed
+  with that subtree when the subtree was later torn down. Two components at the same unkeyed position
+  in *different* containers still share one instance, which is a separate defect; what changes for them
+  is which of the two containers goes stale, since the shared instance's slot index already named the
+  last one reconciled and its container now agrees. Give each an explicit `key:` to keep them apart.
+- A `V.Component` written as a top-level child of a `V.Portal` now survives a patch of that portal's
+  children. Its mount ran in the deferred pass that follows the reconcile, which registered it under a
+  different parent from the one every later patch looks it up by, so the first patch failed to find it:
+  the component was built a second time with fresh hook state and fresh effects, the first instance's
+  effect cleanups never ran, and the element it had rendered stayed behind in the target. A modal whose
+  content component held its own `UseState` reset it whenever anything else in that portal re-rendered.
 - Registering a portal target id again with a different element now moves the portals already mounted
   into the old one, instead of leaving them writing into an element the UI has replaced. A
   `"modal-root"` that a screen owns — torn down on navigation and re-registered from the rebuilt
--- a/Packages/com.velvet.core/CHANGELOG.md
+++ b/Packages/com.velvet.core/CHANGELOG.md
@@ -30,6 +30,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Moving a component across a `V.Portal`'s boundary is an unmount and a remount in every case now, so
+  its state, refs and effects do not survive the move and the departing instance's cleanups run. A
+  component written into a live portal's children that a previous render had outside them — and the same
+  move back out — used to keep its instance where the portal's children were reached by a patch, while
+  the same edit against a portal that mounts or closes in that render mounted a fresh one. Which of the
+  two a given edit got was decided by whether the portal already existed, and neither the guide nor
+  `createPortal` promised the surviving half. The portals guide states the contract that now holds in
+  both directions. Keeping state across such a move means lifting it above the portal — to a `Store`, or
+  to a `UseState` in the component that declares the portal.
 - `V.TextField`'s four new parameters sit between `isPasswordField:` and `enabled:`, so a call passing
   arguments positionally past `isPasswordField` now binds them to different parameters. Every argument
   there but a `null` literal changes type across the shift, so such a call fails to compile rather than
@@ -41,23 +50,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   it is in. Its slot index and the stamp a portal teardown disposes by both followed the move and its
   container did not, so its own `setState` reconciled its new output into the container it had left —
   leaving the stale element behind there and writing the new one into a container it no longer occupies.
-  It also let a component that had moved out of a subtree be disposed with that subtree when the subtree
-  was later torn down. Two components at the same unkeyed position in *different* containers still share
-  one instance where both containers sit in the declaring component's own tree, which is a separate
-  defect; what changes for them is which of the two containers goes stale, since the shared instance's
-  slot index already named the last one reconciled and its container now agrees. Give each an explicit
-  `key:` to keep them apart.
-- A `V.Component` written as a top-level child of a `V.Portal` now survives a patch of that portal's
-  children. Its mount ran in the deferred pass that follows the reconcile, which registered it under a
-  different parent from the one every later patch looks it up by, so the first patch failed to find it:
-  the component was built a second time with fresh hook state and fresh effects, the first instance's
-  effect cleanups never ran, and the element it had rendered stayed behind in the target. A modal whose
-  content component held its own `UseState` lost it the first time anything else in that portal
-  re-rendered — once, because the instance built to replace it registers where the patch looks, so it
-  survives every patch after that. The element the first instance rendered stayed in the target for the
-  life of the tree. Where the declaring component renders the same component at the same position both
-  inside the portal and outside it, the two no longer end up sharing one instance from the first patch
-  onwards: the portal's child keeps the instance it mounted with.
+  Two components at the same unkeyed position in *different* containers still share one instance where
+  both containers sit in the declaring component's own tree, which is a separate defect; what changes
+  for them is which of the two containers goes stale, since the shared instance's slot index already
+  named the last one reconciled and its container now agrees. Give each an explicit `key:` to keep them
+  apart.
+- A `V.Component` written inside a `V.Portal` now survives a patch of that portal's children. Its mount
+  ran in the deferred pass that follows the reconcile, which registered it under a different parent from
+  the one every later patch looks it up by, so the first patch failed to find it: the component was
+  built a second time with fresh hook state and fresh effects, the first instance's effect cleanups
+  never ran, and the element it had rendered stayed behind in the target. A modal whose content
+  component held its own `UseState` lost it the first time anything else in that portal re-rendered —
+  once, because the instance built to replace it registers where the patch looks, so it survives every
+  patch after that. The element the first instance rendered stayed in the target for the life of the
+  tree.
+- A component a declaring render writes both inside a `V.Portal` and at the same position outside it now
+  keeps its own instance on each side, whichever of the two mounts first. They used to merge into one, in
+  both orders and each with its own damage. With the one outside mounted first, the first patch of the
+  portal's children replaced the portal's child with it. With the portal's mounted first, the position
+  outside took the portal's own instance over and rendered it in both places, and the next patch of that
+  portal built a second child and left the elements of the previous slot range in the target with nothing
+  to reconcile them away.
 - Registering a portal target id again with a different element now moves the portals already mounted
   into the old one, instead of leaving them writing into an element the UI has replaced. A
   `"modal-root"` that a screen owns — torn down on navigation and re-registered from the rebuilt
--- a/Packages/com.velvet.core/CHANGELOG.md
+++ b/Packages/com.velvet.core/CHANGELOG.md
@@ -38,19 +38,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - A component the reconciler carries to a different container now re-renders itself into the container
-  it is in. Only its slot index followed the move, so its own `setState` reconciled its new output into
-  the container it had left — leaving the stale element behind there and writing the new one into a
-  container it no longer occupies. It also let a component that had moved out of a subtree be disposed
-  with that subtree when the subtree was later torn down. Two components at the same unkeyed position
-  in *different* containers still share one instance, which is a separate defect; what changes for them
-  is which of the two containers goes stale, since the shared instance's slot index already named the
-  last one reconciled and its container now agrees. Give each an explicit `key:` to keep them apart.
+  it is in. Its slot index and the stamp a portal teardown disposes by both followed the move and its
+  container did not, so its own `setState` reconciled its new output into the container it had left —
+  leaving the stale element behind there and writing the new one into a container it no longer occupies.
+  It also let a component that had moved out of a subtree be disposed with that subtree when the subtree
+  was later torn down. Two components at the same unkeyed position in *different* containers still share
+  one instance where both containers sit in the declaring component's own tree, which is a separate
+  defect; what changes for them is which of the two containers goes stale, since the shared instance's
+  slot index already named the last one reconciled and its container now agrees. Give each an explicit
+  `key:` to keep them apart.
 - A `V.Component` written as a top-level child of a `V.Portal` now survives a patch of that portal's
   children. Its mount ran in the deferred pass that follows the reconcile, which registered it under a
   different parent from the one every later patch looks it up by, so the first patch failed to find it:
   the component was built a second time with fresh hook state and fresh effects, the first instance's
   effect cleanups never ran, and the element it had rendered stayed behind in the target. A modal whose
-  content component held its own `UseState` reset it whenever anything else in that portal re-rendered.
+  content component held its own `UseState` lost it the first time anything else in that portal
+  re-rendered — once, because the instance built to replace it registers where the patch looks, so it
+  survives every patch after that. The element the first instance rendered stayed in the target for the
+  life of the tree. Where the declaring component renders the same component at the same position both
+  inside the portal and outside it, the two no longer end up sharing one instance from the first patch
+  onwards: the portal's child keeps the instance it mounted with.
 - Registering a portal target id again with a different element now moves the portals already mounted
   into the old one, instead of leaving them writing into an element the UI has replaced. A
   `"modal-root"` that a screen owns — torn down on navigation and re-registered from the rebuilt
--- a/Packages/com.velvet.core/CHANGELOG.md
+++ b/Packages/com.velvet.core/CHANGELOG.md
@@ -49,8 +49,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - A component the reconciler carries to a different container now re-renders itself into the container
   it is in. Its slot index and the stamp a portal teardown disposes by both followed the move and its
-  container did not, so its own `setState` reconciled its new output into the container it had left —
-  leaving the stale element behind there and writing the new one into a container it no longer occupies.
+  container did not, so its own `setState` reconciled its new output into the container it had left,
+  while the container it had moved into kept the element from before the move.
   It also let the container it had left take it down: where the arriving container is reconciled first
   and the departing one leaves the tree in that same render, the departing subtree's teardown still
   found the carried component named as living inside it and ran its cleanups.
--- a/Packages/com.velvet.core/CHANGELOG.md
+++ b/Packages/com.velvet.core/CHANGELOG.md
@@ -33,12 +33,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Moving a component across a `V.Portal`'s boundary is an unmount and a remount in every case now, so
   its state, refs and effects do not survive the move and the departing instance's cleanups run. A
   component written into a live portal's children that a previous render had outside them — and the same
-  move back out — used to keep its instance where the portal's children were reached by a patch, while
-  the same edit against a portal that mounts or closes in that render mounted a fresh one. Which of the
-  two a given edit got was decided by whether the portal already existed, and neither the guide nor
-  `createPortal` promised the surviving half. The portals guide states the contract that now holds in
-  both directions. Keeping state across such a move means lifting it above the portal — to a `Store`, or
-  to a `UseState` in the component that declares the portal.
+  move back out — used to keep its instance whenever a patch of the portal's children had been the last
+  thing to register the portal-side occurrence, the render that closes the portal included; only an
+  occurrence the portal's own deferred mount had registered mounted fresh. Which of the two a given edit
+  got was decided by which of those two entrances had last put the child in the portal, and neither the
+  guide nor `createPortal` promised the surviving half. The portals guide states the contract that now
+  holds in both directions. Keeping state across such a move means lifting it above the portal — to a
+  `Store`, or to a `UseState` in the component that declares the portal.
 - `V.TextField`'s four new parameters sit between `isPasswordField:` and `enabled:`, so a call passing
   arguments positionally past `isPasswordField` now binds them to different parameters. Every argument
   there but a `null` literal changes type across the shift, so such a call fails to compile rather than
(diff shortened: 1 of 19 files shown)


s4k10503:
Review round. HIGH 1, MEDIUM 2, MEDIUM 3, MEDIUM 4 and both LOWs are addressed. One sub-claim of MEDIUM 2 did not reproduce and is reported below rather than worked around.

## HIGH 1 — the CHANGELOG sentence

"reset it whenever anything else in that portal re-rendered" is replaced by the measured account: it is lost the first time, once, because the instance built to replace it registers where the patch looks; the element the first instance rendered stays in the target for the life of the tree.

`Given_APortalChildComponentHoldingState_When_ThePortalsChildrenPatch_Then_ItKeepsThatState` now patches twice and is renamed to `..._When_ThePortalsChildrenPatchTwice_...`. On the merge base:

```
Given_APortalChildComponentHoldingState_When_ThePortalsChildrenPatchTwice_Then_ItKeepsThatState
Expected: (7, content|sib-2)  But was: (0, content|sib-2|sib-0)
```

## MEDIUM 2 — the mount-time regression

Reproduced. `V.Div(name: "outside", [V.Component(Twin), V.Portal("t", [V.Component(Twin), V.Div(name: "sib-" + n)])])`, where `Twin` holds a `UseState` and a mount `UseLayoutEffect`:

on this branch before the fix
```
MOUNT: setups=0 marks=2 distinctSetters=1 outside=[twin-a|] target=[twin-a|sib-0]
SET0:  outside=[twin-a|] target=[twin-a|sib-0]
```
on `origin/main`
```
MOUNT: setups=2 marks=2 distinctSetters=2 outside=[twin-a|] target=[twin-a|sib-0]
SET0:  outside=[twin-b|] target=[twin-a|sib-0]
```

One fiber where the base had two — and a symptom the finding did not name: that fiber's mount `UseLayoutEffect` never ran at all, so a user of this shape would get no mount effect and no cleanup rather than a shared one.

**Fix.** The inline registry key gains a fourth member, a portal scope, null everywhere except the one shape that needs it: a portal's own top-level child whose `(parentFiber, positionKey, identity)` is already held by a fiber the portal did not put there. `ReconcilerContext.PortalChildKeyScope` carries the placeholder; `DrainPendingPortalMounts` and `PatchPortalChildren` set it around their nested reconcile and `ExpandComponentInline` clears it before descending, so it names that one level and nothing deeper — a fiber created by a portal child's own isolated re-render carries no stamp, and would otherwise be duplicated on the next patch. `ComponentRegistry.ResolveInline` asks the scoped key first and adopts a plain-key occupant only when this same portal registered it.

Scoping a portal's children unconditionally was tried on paper and rejected: it changes the key of every portal child, so the carry `Given_AFiberCarriedToAnotherContainer_...` pins would become a rebuild.

RED on this branch with only the registry change reverted:

```
Given_OneComponentBothInAPortalAndOutsideIt_When_TheyMount_Then_EachRunsItsOwnMountEffect
Expected: (2, twin-a|sib)  But was: (0, twin-a|sib)

Given_OneComponentBothInAPortalAndOutsideIt_When_TheOutsideOneSetsItsState_Then_ThePortalsKeepsItsOwn
Expected: (twin-b, twin-a|sib)  But was: (twin-a, twin-a|sib)
```

Both carry `GREEN_ON_BASE(characterization)`: the base separates the two occurrences by accident, the drained one landing on the reconcile root.

Mutating the adoption test (`ReferenceEquals(plain.OwningPortalPlaceholder, portalScope)` to `false`) reddens four of the fixture's other cases, so the discriminator is pinned by cases already here.

**What did not reproduce.** The finding says closing the portal disposes the in-tree component, `CleanupPortal` → `DisposeInlineFibersOwnedByPortal(placeholder)` selecting on the stamp. It does not, in this shape: the close is a re-render of the declaring component, whose own walk re-claims the shared fiber and rewrites `OwningPortalPlaceholder` before the teardown selects. Measured on this branch without the fix, after the close the element outside is still present and still re-renders on its own `setState`. A case written for it passed without the fix, so it was deleted rather than shipped vacuous.

**What did not change.** Moving a component *into* an existing portal's children is an unmount and a remount, and reads identically on `origin/main` and here: `setups 1 -> 2`, state back to its initial value, `outside=[] target=[rover-a]`. The refusal to adopt across that direction is not new.

## MEDIUM 3 — the two surviving mutants

`Given_APortalBesideAContextConsumingSibling_When_TheSiblingReRendersAlone_Then_ItStillReadsTheProviderAroundIt`, `GREEN_ON_BASE(characterization)`. Each mutant applied alone, whole fixture run:

- `before.Add(f)` removed — `Expected: (inside, |sibling-1)  But was: (outside, |sibling-1)`
- `if (before!.Contains(f)) continue;` removed — the same reading.

Every other case in the fixture stayed green under both.

## MEDIUM 4 and the LOWs

- `FiberNodeFactory.EnqueueDeferredHostMount` no longer claims nothing on the stack reflects the declaring fiber by drain time; it says the capture is what makes it available and that the drain pushes it back.
- The fixture's class doc now names which move is an unmount and a remount (a surviving portal re-homing its child, or a re-registered target id) and which is a carry (leaving the children in the render that removes the portal).
- CHANGELOG: "Only its slot index followed the move" becomes the slot index *and* the stamp a portal teardown disposes by, with the container the one that did not.
- Found in passing and fixed: `PortalContextInheritanceTests`' `<remarks>` still said the drain parents portal children off the reconcile root.

No `Documentation~` change: `portals.md` states the logical-tree contract and the retarget contract, and neither moves. The portal CHANGELOG bullet gains one sentence for what does change against the released version — a component rendered at one position both inside a portal and outside it no longer ends up sharing a single instance from the first patch onwards.

## Suites

EditMode 3888 passed / 0 failed / 0 inconclusive. PlayMode 161 passed / 0 failed / 0 inconclusive. `assert_no_inconclusive.py` clean on both; every run log checked with `grep -c "error CS"` before its results file was read. `base_red_check.py --lane csharp --base origin/main` exits 0: four cases red on the base, four declared and green as declared.

s4k10503:
## The design changed, and one premise behind the change turned out to be half wrong

A portal's children are now scoped **unconditionally**: the portal placeholder is a member of the
inline registry key, not a fallback the lookup tries second. `ComponentRegistry.ResolveInline` is
gone; the key is `(parentFiber, portalScope, positionKey, identity)` and nothing adopts anything.

### Finding 1 is closed by it

Measured on this branch at `31d11d1`, with the portal's occurrence mounting first and an unkeyed
in-tree occurrence arriving later (`r<n>` is the instance, `psib<n>` a portal sibling):

```
p0  target=[r1a,psib0]  outside=[_]        setups=1 ids=1
p1  target=[r1a,psib1]  outside=[_,r1a]    setups=1 ids=1     <- one instance in both places
p2  target=[r2a,psib2,psib1]  outside=[_,r1a]  setups=2 ids=2 <- second child, psib1 orphaned
```

With unconditional scoping:

```
p0  target=[r1a,psib0]  outside=[_]        setups=1 ids=1
p1  target=[r1a,psib1]  outside=[_,r2a]    setups=2 ids=2
p2  target=[r1a,psib2]  outside=[_,r2a]    setups=2 ids=2
```

`origin/main` is broken here too, differently: `p1 target=[r2a,psib1,psib0] outside=[_,r2a]` — one
instance in both places and `psib0` orphaned from the first patch. Both orderings are now pinned.

### The carry: the review was right about the case, and wrong about the behaviour

`Given_AFiberCarriedToAnotherContainer_…` is genuinely **red on `origin/main`** —
`base_red_check.py` reads `Expected: (b, c, )  But was: (a, c, )`. The mark is back at `a` there, so
the mount-shape carry the case was written for does not exist on the base, exactly as round 3 said.

But a **different shape of the same carry is `origin/main` behaviour, and `origin/main` pins it**.
`PortalRegistryRetargetTests.Given_AComponentReHomedInThePortalsClosingRender_When_ThatRenderCommits_Then_ItSurvivesTheClose`
asserts `s_childCleanups == 0` for a component that leaves a portal's children — and it passes on
`origin/main` (22/22 there). Its own fixture comment says why: that component is written into the
portal's children **by a patch**, and the patch entrance has always reconciled a portal's children
with the declaring fiber current, so the two positions already shared a whole key.

So `origin/main`'s behaviour is: a component crossing the portal boundary keeps its instance when
the portal's children were last reached by a patch, and mounts fresh when the same edit mounts or
closes the portal in that render. Nothing documents that split and no reader could predict it. This
change makes both a fresh mount, which is what React does and what `Documentation~/portals.md` says
of a move. That test is rewritten to the question it was actually asked to answer — whether the
close reaches past its own children — and the state-carry term is dropped from it.

**This is a user-visible behaviour change against the released version and it has its own
`### Changed` entry.** If keeping the patch-shape carry matters more than the parity, this is the
commit to argue with: the two cannot both hold, because the carry exists only while the two
positions share a key, and sharing a key is what finding 1 is.

### Finding 3

The comment on `ReconcilerContext.PortalChildKeyScope` claimed a fiber deeper in the portal never
sees the scope. It does, whenever only plain host elements sit between — the code was right. The
comment now says how far the scope reaches and why the clear sits at the component boundary, and a
case holds the two registry keys of `V.Portal(id, [V.Div([V.Component(Twin)])])` and
`V.Div([V.Component(Twin)])` side by side: equal in parent, position key and identity, differing in
the scope alone.

### Also found while measuring

- `FiberContextSpine.MatchesInlineSpineChild` passed `SpineChild.OwningPortalPlaceholder` as the
  scope. That is the placeholder for **every** fiber in a portal's subtree, including ones registered
  with no scope at all — a consumer a component level below the portal's children. It worked on the
  old branch only because the plain-key fallback rescued it. With the fallback gone it loses the
  Provider: neutering the fix reads `seen=outside` after the isolated re-render, while all four
  `PortalContextInheritanceTests` cases stay green, because their consumers are the portal's own
  children. The spine now reads the scope back off the registration.
- The `[Unreleased]` entry claiming a carried component could be disposed with a subtree it had left
  is dropped: a component moved between two sibling containers is not carried at all (it is disposed
  and rebuilt at the new position), so nothing measured supports it. What is left of that entry is
  pinned by the two-container shape, where the instance genuinely is shared.

EditMode 3895 passed / 0 failed / 0 inconclusive; PlayMode 161 passed / 0 failed / 0 inconclusive.
`base_red_check.py --lane csharp --base origin/main` exits 0.

s4k10503:
## Review round: the scope was ambient over renders that never repeat under it

### HIGH — fixed by moving the limit off the clear site

`PortalChildKeyScope` was set around the whole nested reconcile and cleared in one place,
`GeneralPathReconciler.ExpandComponentInline`. Enumerating the entrances that render a tree under the
declaring fiber gives four `FiberStack.Push` sites and one detached-item scope; the clear covered one of
them. The two that repeat their render away from the portal are the ones that break:

- **VirtualList items.** `PatchPortalChildren` → `PatchNode` → `controller.Update` → `RenderRange`. The
  geometry-driven render keys with no scope, the patch-driven one with the placeholder.
- **The wrapper-mount entrance.** An `V.Outlet()` among a portal's children mounts its route Component in
  the drain, so its body's inline children register with the scope; every later render is
  `RequestRenderFromHook`, with none.

Rather than adding a clear per entrance, the scope now carries the **render nesting** it was entered at —
`FiberStack.Depth` plus a new `ReconcilerContext.DetachedItemScopeDepth` — and `PortalChildKeyScopeHere`
answers only there. Each of those entrances is reached by pushing one of the two counters, so each stops
the scope without naming it, and `ExpandComponentInline`'s clear and its `try`/`finally` go away. The
detached-item counter is bumped **ahead of** that scope's `host == null` gate, which is the one crossing a
fiber-stack depth alone cannot see.

Both cases red on `77372de`, tests unchanged, production reverted by patch:

```
Given_AComponentInsideAVirtualListInAPortal_..._TheItemKeepsTheInstanceItsGeometryRenderBuilt
  Expected: (item-1, row-b)   But was: (item-1, row-a|row-b)
Given_AnOutletRouteInsideAPortal_..._ItsOwnChildKeepsItsInstance
  Expected: (1, inner-b)      But was: (2, inner-a|inner-b)
```

Both carry `GREEN_ON_BASE(characterization)`: `main` has one key either way, so it agrees for want of a
second key rather than by anything holding it to one.

`V.AnimatePresence` children are confirmed unaffected — `EmitPresenceChild` expands them through
`ExpandInlineRecursive`. The `FiberNodeFactory` comment claiming that shape wrapper-mounts was stale and
is corrected; enumerating all twelve `CreateElement` call sites, the only one handed a node the walk never
expanded is `IReconcilerBridge.CreateElementForController`.

### MEDIUM-HIGH — the `### Changed` entry named the wrong discriminator

The entry said a portal that **closes** in that render mounted fresh. It did not, where the child had been
patched in: `PortalRegistryRetargetTests.Given_AComponentWrittenOutsideAPortalInItsClosingRender_...`
drives phase 1 → 3 and reads green on the base. What decided it was which entrance had last registered the
portal-side occurrence — the deferred mount, or a patch of the portal's children — and the entry says that
now.

### MEDIUM — the guide's comparison is gone, and a deleted claim is back

`portals.md` no longer compares a boundary crossing to "moving it between two ordinary parents": those two
share a whole registry key, so whether such a move remounts depends on which container the reconcile
reaches first.

Re-measured in both orderings. Where the **arriving** container is reconciled first and the departing one
leaves the tree in the same render, the departing subtree's teardown reaches the carried component and runs
its cleanups — `DisposeFibersUnder` selects by `MountPoint`, which did not follow the carry. A new case
reads `Expected: (0, b)  But was: (1, b)` on `origin/main`, so the sentence about a carried component
disposed with a subtree it had left is restored to the `### Fixed` entry. The previous measurement covered
only left→right, where the departing container's own orphan sweep disposes and rebuilds anyway.

### LOW — the placement stamp

`existingFiber.OwningPortalPlaceholder = _ctx.CurrentPortalPlaceholder` now carries a comment saying what
it does rather than being read as dead: it is rewritten even where the reconcile names no portal, which
erases the stamp a fiber below a portal's own child was given at creation. Nothing follows from that —
`DisposeInlineFibersOwnedByPortal` reaches the same fiber through the parent index.

### Verification

EditMode 3898 passed / 0 failed / 0 inconclusive. PlayMode 161 passed / 0 failed / 0 inconclusive.
`assert_no_inconclusive.py` clean on both; `grep -c "error CS"` zero on every log.
`base_red_check.py --lane csharp --base origin/main` exits 0 over 20 cases — 9 red on the base, 11 declared
and green as declared. Generator suite 422 passed / 0 failed.

No new CHANGELOG entry is owed for the scope fix itself: the defect it repairs exists only on this
unmerged branch, and the `### Changed` / `### Fixed` entries already describe what changes against the last
release.

Issue #448, opened 2026-08-06
Thirteen class-driven mechanisms have no neuter cut, and no workflow runs the sweep


s4k10503:
> **Re-measured against `origin/main` (`89042a0`) on 2026-08-13.** The first half of this issue — the
> decorative name-anchored scan — is closed: `NeuterCutRegistrationDriftTests`, `GuardFixtures()` and
> `RegistrationRequiredPatterns` no longer exist anywhere in the tree. What replaced them derives the
> answer from the mechanisms rather than from fixture names, and its own docstring names this issue's
> finding as the reason. The other two halves stand.

## What is closed

`NeuterCoverageTests` (`Packages/com.velvet.core/Runtime/Component/Tests/Editor/NeuterCoverageTests.cs`)
replaced the registration guard. It globs every `Runtime/Styling/Style*Class.cs` and
`Runtime/Reconciler/Fiber*Applier.cs` (30 files at this commit), subtracts the files
`neuter_cuts.json` edits, and diffs the remainder **both directions** against a checked-in record —
so a mechanism added tomorrow must be either cut or recorded, and a cut somebody writes must leave
the record. A second case fails when a recorded entry names a file a rename removed.

Its docstring also records why deriving the requirement from fixture *text* was tried and abandoned:
a cut edits `FiberNodePatcher`, so ten unrelated fixtures mentioning it read as required.

## What remains

**Thirteen mechanisms still have no cut.** `scripts/test_quality/neuter_coverage.txt` is the
*uncovered* set (`NeuterCoverageTests.cs:103-105` subtracts the cut files and compares the
remainder), and it holds 13 of the 30:

- appliers: `FiberAnimateMotionApplier`, `FiberFilterTransitionApplier`, `FiberGestureApplier`,
  `FiberPropApplier`
- parsers: `StyleAttributeVariantClass`, `StyleChildVariantClass`, `StyleFontClass`,
  `StyleHasVariantClass`, `StyleStructuralVariantClass`, `StyleSupportsVariantClass`,
  `StyleTextBalanceClass`, `StyleTextEffectClass`, `StyleVariantClass`

Every one is a C#-realised styling mechanism whose failure mode is a silent no-op — which is what the
harness exists for, and which two defects filed the same week (#392, #394) turned out to be.
Recording a hole is not closing it; the record makes the hole visible and answers for new arrivals,
and that is all it does.

**No workflow runs the sweep.** `git grep neuter origin/main -- .github/` returns exactly one line,
[...]

PR #640, opened 2026-08-13
canary: show the neuter audit failing in CI

s4k10503:
Refs #448. Not for merge — closed as soon as the run reports.

This branch removes the `has-parser` cut and its fixture registration from `neuter_cuts.json`, which
leaves `Runtime/Styling/StyleHasVariantClass.cs` a class-driven mechanism nothing can disable and
nothing records. It exists to show `Test ▸ test-quality` going red for that, since a guard nobody has
watched fail has not been shown to work.


s4k10503:
Its purpose is served: `Test quality` step 5 exited 1 with `Runtime/Styling/StyleHasVariantClass.cs has no cut and is not recorded`, which is the reading the audit exists to produce.

Issue #635, opened 2026-08-13
The verdict lane gets no Unity log, so a compile failure the static reading cannot attribute reports no cause


s4k10503:
`base_red_check` answers a different question locally than it does on CI, and the local answer is the
permissive one. A branch can read green on a developer's machine and red on the pull request, for a
reason the developer cannot reproduce.

### Measured

On a branch whose new guard fixture named API that does not exist on the base, the local run printed:

```
Aborting batchmode due to failure:
Scripts have compiler errors.
EditMode attempt 1: 0 case(s) over 5 fixture(s) in 19s
EditMode attempt 2: 26 case(s) over 4 fixture(s) in 54s
```

Attempt 1 hit the compile abort and answered nothing. The script then **withdrew the uncompilable
file and repeated**, and the second attempt answered 26 cases across the four remaining fixtures —
six of them red on the base. Exit 0.

The same branch on CI:

```
the base run wrote no result, so nothing it was asked was measured
the base tree cannot answer   <14 cases>   (the run wrote no results file, so nothing was measured)
```

Exit 1.

### Why

The two are different flows. Locally the script drives its own rounds and honours `--max-rounds`
(default 8), withdrawing a carried file the base cannot build and asking again. CI runs
`--emit` → one Unity run → `--verdict`, which its own docstring describes as "one round of it,
without the withdrawing." One uncompilable carried file therefore blinds the entire base-red check for
that branch on CI, while locally it costs only that file.

**A warm `Library` is not the cause.** `clone_tree` does copy the whole `Library` including
`ScriptAssemblies`, and the seeded assemblies do carry the branch's API — but sources decide the
compile, and the sources are the base's. Removing the `Library` cannot make a compile succeed that
failed with it. This is a different gap from #608's, and should not be recorded as an instance of it.

### Which answer is right

CI's refusal is correct for the cases in the uncompilable file: the base could not be asked about
them. It is **not** correct for the other twenty-six, which the base can answer and did answer
locally. So CI is not merely stricter — it discards readings that exist, and it discards them
because of an unrelated file.

The local flow is not right either, in the direction that matters here: it exits 0 while a carried
file went unasked, so a developer's green does not mean what the pull request's green means.

### What a fix has to reconcile

- CI must be able to withdraw an uncompilable carried file and ask again, or say clearly that it
  cannot and that the branch has to make the file base-buildable first. Today it says neither: the
  message names the fourteen cases and not the one file that caused it.
[...]

PR #670, opened 2026-08-15
fix(velvet): resume the navigation a Blocker proceeded, and bound the fixtures that wait on one

### LINKED ISSUE #416: Blocker.Proceed() does not resume the blocked navigation
## What React does

`useBlocker` (v6.7+) returns a blocker whose `proceed()` retries the exact navigation that was blocked — history action, target location and all — moving it through `"proceeding"` to completion. `reset()` cancels it. The caller never re-issues the navigation.

## What Velvet does

`RouteBlockerState.Proceed()` flips Status to Idle, clears `Attempt`, and invokes `OnProceed` — a field that nothing in `Runtime/` ever assigns. The XML doc states it: "After calling Proceed() you must invoke NavigateAsync manually again." The `Attempt` (the only record of where the user was going) is nulled by the same call.

## What a developer hits

The canonical unsaved-changes dialog is broken by default: the user clicks "Leave", the dialog closes, and nothing happens. Correct use requires snapshotting `state.Attempt.NextPath`/`NavigationMode` *before* calling Proceed and re-issuing it, and a Back/Forward attempt cannot simply be replayed through `NavigateAsync(path, mode)` — it needs `GoBack()`/`GoForward()` to keep the history index consistent.

## Where

/Users/apple/dev/velvet/Packages/com.velvet.core/Runtime/Routing/RouteBlockerState.cs, /Users/apple/dev/velvet/Packages/com.velvet.core/Runtime/Routing/RouteBlockerManager.cs

Confirmed by reading the sources; the claim is accurate on every point.

1. `OnProceed` is dead. A grep for `OnProceed`/`OnReset` over the whole of `Packages/` returns exactly four lines, all in `/Users/apple/dev/velvet/Packages/com.velvet.core/Runtime/Routing/RouteBlockerState.cs` — the two `internal Action?` declarations (lines 26, 29) and the two reads (`public void Proceed() => Release(OnProceed);` line 35, `Reset()` line 38). Nothing assigns them: not `Hooks.UseBlockerCore` (Hooks.cs:483-525, which constructs `new RouteBlockerState()` at line 515 and sets only `State`/`NextDeps`/`NextNeedsReregister`/`NextRegister`), not `RouteBlockerManager.Register` (which stores the state on a `BlockerEntry` untouched), not `Router`. So `Proceed()` invokes a null delegate.

2. Nothing retains the blocked navigation for a retry. `Router.RunBlockerCheck` (Router.cs:462-494) constructs the `NavigationAttempt` locally at line 469, passes it to `CheckAsync`, and on `blocked` does only `_historyIndex = savedHistoryIndex; Status = RouterStatus.Idle; return NavigationResult.Blocked;`. The attempt is never stashed on the router. The only surviving copy is `RouteBlockerState.Attempt`, and `Release` nulls it (line 48) *before* invoking the callback — so even if `OnProceed` were wired, it would fire with `Attempt` already null.

3. There is no `"proceeding"` state. `RouteBlockerStatus` (RouteBlockerState.cs:6-12) has exactly two members, `Idle` and `Blocked`. React's blocker state machine is `unblocked | blocked | proceeding`, and the third exists precisely so the retry is not re-vetoed.

4. Worse than "re-issue it yourself": the manual re-issue is re-blocked. `RunBlockerCheck` calls `_blockerManager.ResetAllBlocked()` and then `CheckAsync` on *every* attempt, so a manual `NavigateAsync` after `Proceed()` re-runs the same predicate and blocks again. The caller must make the predicate self-disarm. The fixture that covers this says so — `Given_PreviouslyBlockedNavigation_When_NavigatingAgain_Then_ResetsAndCommits` (Runtime/Routing/Tests/Editor/BlockerTests.cs:311) registers `_ => ++blockCount == 1` with the comment "The blocker blocks only its first invocation". React's `proceed()` does not re-consult the blocker.

5. No test exercises `Proceed()` at all. `BlockerTests.cs` is the only Blocker fixture and none of its ~30 tests calls it; a repo-wide grep for `.Proceed()` outside `docs/_site` finds no call site in `Packages/` or `Assets/`.

6. The deviation is stated but not justified, so it does not qualify as a documented deliberate deviation. RouteBlockerState.cs:33 says "After calling Proceed() you must invoke NavigateAsync manually again" — that is a restatement of the behavior with no reason. The one nearby "by design" comment covers a *different* deviation (Router.cs:470-471: "By design, a different navigation lifts the previous block even if Proceed() was not called (intentional)"). `Documentation~/` has no routing guide and no occurrence of "blocker"; `react-migration.md:559` lists "Routing features (equivalent to React Router, not yet documented separately)" as an undocumented area, and CHANGELOG has no "proceed" entry.

Scope note on what is *not* a gap: `Reset()` is adequate parity. By the time a UI can call it, `RunBlockerCheck` has already rolled back `_historyIndex`, set `Status = RouterStatus.Idle` and returned `NavigationResult.Blocked` — the navigation is genuinely cancelled, so `Reset()` only has to clear the dialog state, which it does. The gap is `Proceed()` alone.

## What exists today

`Hooks.UseBlocker` (Hooks.cs:451-527, sync and async `UniTask` overloads) returns a shared `RouteBlockerState` with `Status` (`Idle`/`Blocked`) and `Attempt` (`NavigationAttempt`: `CurrentPath`, `NextPath`, `NavigationMode`), registered through `Router.RouteBlockerManager`. That covers React's *detection* half — a blocked navigation surfaces as `NavigationResult.Blocked` with the target visible on `Attempt`, enough to drive a confirm dialog.

The *resolution* half is where it stops. `Proceed()` is a state-flip only: it sets `Status = Idle`, nulls `Attempt`, and calls a never-assigned `OnProceed`. To actually continue, the caller must capture `Attempt.NextPath`/`NavigationMode` before calling `Proceed()`, arrange for its own predicate to stop blocking (otherwise the retry is vetoed again), and then re-issue `router.NavigateAsync(path)` / `GoBack()` / `GoForward()` itself — reconstructing by hand what React's `proceed()` does internally. The `OnProceed`/`OnReset` fields look like the intended seam for wiring this up and were left unconnected.

s4k10503:
Closes #416.

`RouteBlockerState.Proceed()` released the block and invoked a callback nothing ever assigned, so the
navigation the Blocker was holding was simply gone. The confirm-dialog flow the API exists for — a dirty
form vetoes a departure, the user clicks "Leave" — left the router where it was. Reaching the destination
meant reading `Attempt.NextPath` and `Attempt.NavigationMode` back out, re-issuing them by hand, and
arranging for the predicate to stop blocking first, because the re-issued attempt was put to that
predicate again.

`Proceed()` now re-issues the navigation, and the pipeline runs again from matching without consulting
the Blocker that released it.

## What is re-issued

The request the caller made — its path and its mode — rather than the attempt the Blocker was handed.
The two come apart once a Guard has rewritten the destination: a Guard rewrites the path, and rewrites a
Back or Forward into a Replace, so the attempt no longer says which history slot it belongs in.
Re-issuing the caller's request means a blocked Back resumes as the same step, and a Guard redirect over
it lands on the slot that step resolves instead of overwriting the entry the user is standing on.
`Attempt` still describes where the navigation was heading when the predicate saw it, which is what a
dialog names the destination from.

## A third state

`RouteBlockerStatus.Proceeding` covers the span between `Proceed()` and the re-issued navigation
settling. Over it the Blocker still reports its `Attempt` and is consulted about nothing, so a predicate
that still answers "block" does not have to disarm itself. It returns to `Idle` — which is what arms it
for the next navigation — once that navigation has committed, ended without committing or been
abandoned, and no Blocker is left `Blocked`. On a landing it is already `Idle` by the time
`Router.OnLocationChanged` runs, so a handler reading a Blocker off that notification sees one that has
finished proceeding. An exhaustive `switch` expression over `RouteBlockerStatus` needs an arm for the
new member.

## More than one Blocker

Velvet consults every registered Blocker where React Router consults one, so several can veto a single
attempt. The order they are answered in does not matter: each `Proceed()` re-issues the attempt, the
Blockers that have already consented stay out of the way, and the ones that have not veto it again until
none is left blocking. `Reset()` from a Blocker still blocking now ends the attempt for the ones holding
it alongside, so a later `Proceed()` on one that had already consented no longer sends the router at a
destination the user declined.

`Proceed()` and `Reset()` both do nothing unless the Blocker is `Blocked`, so a stale dialog handler is
a no-op.

## Two guards

The uncalled-runtime-member check now covers assembly-visible setters. An unwritten setter is the shape
the unassigned `OnProceed` callback had: the call sites reading a property are not evidence that
anything ever writes it.

The three router fixtures that synchronise on a Blocker being consulted carry a `[Timeout]`. The signal
they await is raised from inside the predicate, so a change that stops Blockers being consulted never
raises it: 16 tests across `RouterTests`, `RouterCancellationUnwindTests` and
`RouterUnfinishedNavigationTests` then hang the run indefinitely instead of failing, which also puts the
consultation gate beyond what a mutation run can measure.

## Documentation

Navigation blocking gets its own guide, `Documentation~/routing-blockers.md`, covering the three states,
what `Proceed()` re-issues and what `Reset()` abandons, when a Blocker is armed again, several Blockers
on one attempt, and where the Blocker phase sits relative to matching and Guards. It is listed in the
`Documentation~/README.md` index table and in `docs/toc.yml`, and `react-migration.md`'s routing line
points at it. The CHANGELOG carries the behaviour change under `[Unreleased]`.


s4k10503:
`Base-red check (EditMode)` is red here for a stable, structural reason rather than a flake. Re-running
the failed jobs reproduced it identically.

```
the base run wrote no result, so nothing it was asked was measured

the base tree cannot answer   Velvet.Tests.BlockerProceedTests.Given_ABlockedPush_...
                              (the run wrote no results file, so nothing was measured)   ×12
```

Every affected case is in `BlockerProceedTests.cs`, which this branch **adds**. `base_red_check` carries
the branch's test-side files onto the base tree, and that file names `RouteBlockerStatus.Proceeding` —
measured, `Proceeding` occurs **zero** times in `Runtime/Routing/*.cs` on `main` and in three files on
this branch. So the base tree cannot compile, writes no results file, and every carried case reads as
"the base could not answer", which is a failing verdict.

### It unblocks when #634 lands, and not before

That is exactly the withdrawal #634 adds: read a carried file the base cannot build, withdraw it, and
ask again. Measured — `unbuildable` occurs **0** times in `scripts/test_quality/base_red_check.py` on
`main` and is present on `fix/readings-nobody-took`.

So there is nothing to fix on this branch. Re-running the job will keep producing this until #634 is on
`main`; after that, the same twelve cases come back with a real base verdict.

Related: this is #635's shape — a branch whose carried file cannot compile on the base reads green
locally and red on CI — seen from the CI side.

PR #763, opened 2026-08-22
tooling(velvet): ask which assertion term depends on a condition the test arranged

s4k10503:
No issue: the pre-report questions audit the assertion a test makes and nothing about the conditions it arranges.

"For each assertion you added, which single cut reddens it" is answerable, honestly and in full, by a
fixture whose arrangement is decoration. The cut reddens the case; the thing the case is named for was
never exercised. The two questions fail apart, and only the first was being asked.

Three that reached review this way:

- a fixture whose `<summary>` says each contained failure reaches the nearest error boundary, in which
  no case registers a boundary — every one expects the log line printed when there is none, which is
  the single path where the boundary code never runs;
- a case named for a task completing off the main thread, with no assertion in it reading which thread
  completed;
- a hand-rolled struct state machine calling a builder method the compiler emits zero calls to, which
  is the call that repairs the defect the fixture exists to catch.

Each answered the cut question. Each is a name asserting something no assertion holds.

The fourth question asks for the disposition of every arranged condition, and names the way to answer
it: invert the condition and run. Register the boundary, complete on the main thread, delete the line.
An assertion that does not move says the condition was decoration.

The list's introduction said each of its members has caught a defect **that shipped** past a fully
green suite. That is true of the three; it is false of this one, whose three were stopped in review.
The property the sentence reaches for holds over all four — a green suite did not see any of them — so
the introduction says that instead, and the count moves from three to four.

No documentation impact: `CLAUDE.md` owns what a test must pin, and this restates none of it — it is a
report requirement, which is what this file owns. No CHANGELOG entry — contributor tooling.


s4k10503:
Withdrawn. A review found three things, and two of them are shapes this repository already names.

**The added text states a fact about somebody else's compiler and gets it wrong.** Measured, with `ikdasm` over both optimization settings: Roslyn emits exactly one call to `Builder.SetStateMachine`, from the state machine's own `SetStateMachine` method. What it emits none of is a call **from `MoveNext`** — which is what the fixture's comment says and what the commit that found the defect says. I dropped the qualifier and the body strengthened it to "emits zero calls to". `CLAUDE.md` names this as the source of every false comment it lists, and it went into the file whose job is to tell implementers not to write them.

**The first example is not the class the bullet names.** `ElementCallbackFailureTests` arranges nothing — no boundary is the default state of a bare reconcile fixture, not a condition anyone set up. Run the bullet's own recipe on it and registering a boundary *moves* the check, so the recipe answers "not decoration". The real class is a summary claiming a route no case exercises, and its remedy is a new case rather than an inversion — which is what `fix/contain-ref-and-create-callbacks` has since done, uncovering a production defect behind it.

**And it restates the `unity-tests` skill**, which `velvet-implementer.md` loads in its own frontmatter. The skill carries the item as the ninth of "Nine ways a test here has passed while checking nothing", using the same word, and carries the remedy in its sweep section. The same file says of the universals list: *do not restate its list here*. The two copies had already drifted at birth — the skill says remove the condition and check that something goes red, this said invert it and check that the assertion moves.

Separately: #744 considered adding a fourth member to that list and declined, on the grounds that the introduction's "each has caught a defect that shipped past a fully green suite" would go false over the widened set. That reasoning is on `main`. This made the move it declined and resolved the falsity by weakening the introduction instead, without saying so — and the weakened claim is close to vacuous, since a suite that had seen a defect would have gone red.

The one example that carried the distinction was the thread pair, and it carries it cleanly.

What is worth keeping is narrower than a fourth question and belongs where #744 put its sibling: a line under `## Reporting` asking for the disposition of each arranged condition, pointing at the skill for how to test one rather than restating it. That is the next change.

PR #824, opened 2026-08-28
tooling(velvet): read a trigger filter written on the trigger's own line

### LINKED ISSUE #737: WorkflowTriggerCoverageTests cannot see a flow-style child key, so a gate re-added in flow form passes silently
Title: WorkflowTriggerCoverageTests cannot see a trigger filter written as a flow mapping

`WorkflowTriggerCoverageTests.TriggerFilters` decides what a child key is from a line-and-indent
match: `KeyPattern` finds a key at the start of a line, the trigger is the one at indent 2, and a
child is anything deeper. A trigger written as a YAML flow mapping puts its child on the trigger's
own line, where `KeyPattern` matches the trigger name and the guard yields nothing.

Measured against the two required workflows as they stand on `2.x`, one line replaced each time:

| workflow line | guard |
|---|---|
| `  pull_request: {branches: [main, 2.x]}` | GREEN `(4, "")` |
| `  pull_request: {paths: ['**.md']}` | GREEN `(4, "")` |
| `  merge_group: {branches: [main]}` | GREEN `(4, "")` |

The block-style equivalents are all red, so this is the spelling and not the key.

Why it matters rather than being a curiosity: the guard exists because a `pull_request` branch
filter stopped both required workflows from starting for a pull request based on anything but
`main`, and no ruleset covers a maintenance branch, so the absence of a check looks like a pull
request that had nothing to run. If that filter is re-added in flow form — by hand, or by a
formatter collapsing a one-entry mapping — the guard reports nothing and the suite stays green.

Yield today is zero: no trigger key in any of the four workflow files carries an inline value,
measured over 13 trigger keys. But `branches: [main]` is already written as a flow *sequence* in
both files, so flow syntax is in-house style rather than exotic, and the collapse is a plausible
edit rather than a hypothetical one.

Remedy, untested: yield the text after a gated trigger's colon as well as the indented children.
The detail that bites is that `KeyPattern` matches a prefix, so a trigger line carrying a value
still matches as a key and the reader has to be taught to look past the colon rather than to stop
there. Worth a case per flow spelling, since each is a different point in the same reader.

s4k10503:
`WorkflowTriggerCoverageTests` refuses a path or branch filter on `pull_request` and `merge_group`,
because a required check that a filter stops from starting stays Pending with nothing able to clear
it. It reads a trigger's children by line and indent, so the one spelling that puts a child on the
trigger's own line goes past it: `pull_request: {branches: [main]}` passed the guard that fails the
block-style spelling of the same filter — and a branch filter is exactly what left a maintenance-line
pull request with no required check at all.

The trigger's own line is now read after its colon, counting only a `{`- or `,`-preceded name, so a
value that happens to hold a colon names no key. Both assertions are unchanged: the gated triggers
must carry no child, and the push floor still rides along.

The case writes a workflow of that shape to a temporary file and reads it back, so what fails when
the reading is removed is the reading rather than a comment about it.

No documentation impact: the guard is a test, and CLAUDE.md describes what it refuses rather than
how it parses.

Closes #737


s4k10503:
Verified RED by hand rather than by the base lane: removing only the `FlowKeys` call from `TriggerFilters` and re-running the fixture gives 7 passed / 1 failed, the failure being this case. The base lane cannot separate them because the reading lives in the fixture it tests.

PR #840, opened 2026-08-28
tooling(velvet): read the unwalked roots off what git tracks, not off the filesystem

### LINKED ISSUE #802: DocumentationDriftTests reddens on any untracked top-level directory holding markdown, so the local answer and the CI answer differ
`DocumentationCorpus.UnwalkedMarkdownRoots()` enumerates the real filesystem:

```csharp
Directory.EnumerateDirectories(Path.GetFullPath("."))
…
if (walked.Contains(name) || BaseUnwalkedDirectories.Contains(name) || ignored.Contains(name)) continue;
```

so any top-level directory holding a `.md` reddens
`Given_EveryTopLevelDirectoryHoldingMarkdown_When_TheWalkIsRead_Then_TheWalkReachesIt` — including one
git has never heard of.

## Measured

The primary checkout holds an untracked `.agents/` with two markdown files
(`skills/release/SKILL.md`, `skills/unity-tests/SKILL.md`). It is in neither `WalkedRoots` nor
`BaseUnwalkedDirectories`, and `.gitignore` names neither `agents` nor `codex`, so the case reports
`(true, ".agents")` against `(true, "")`.

CI checks out tracked files only, so `.agents/` never exists there and the same case is green. A fresh
worktree is green too. The red is machine-local and belongs to whoever happens to have an untracked
directory with a `.md` in it — a scratch note, a vendored tool, an agent harness.

## Why the reading is the defect rather than the directory

The guard exists to catch a documentation directory somebody added and did not add to the corpus. It
cannot tell that apart from an untracked local directory, and the two want opposite answers.

git already draws the line. A new documentation directory is tracked at the moment of the commit that
adds it, which is before CI and before review, so a reading scoped to what git knows about catches
every case the guard is for and none of the case it is not. That also makes the local answer and the
CI answer the same, which `DocumentationCorpus`'s own remarks say the rooting exists to achieve.

## Yield

Replaying is cheap here because the loss is bounded by construction: an untracked-but-intended docs
directory is caught on its first commit. Nothing historical is missed, since every documentation
directory in the repository is tracked.

Found while reviewing #785, whose reviewer hit the red in the primary checkout. Not caused by that PR.

s4k10503:
`DocumentationCorpus.UnwalkedMarkdownRoots` enumerated the filesystem, so any top-level directory
holding a `.md` reddened the guard — including one git has never heard of.

Measured: the primary checkout holds an untracked `.agents/` with two documents. It is in neither
`WalkedRoots` nor `BaseUnwalkedDirectories`, and `.gitignore` names it nowhere, so the case reported
`(true, ".agents")`. CI checks out tracked files only, so the same case is green there, and so is a
fresh worktree. The red is machine-local and belongs to whoever happens to have a scratch note, a
vendored tool or an agent harness sitting at the top level.

The reading is the defect rather than the directory. The guard exists to catch a documentation
directory somebody added and did not add to the corpus, and it could not tell that apart from an
untracked local one — the two want opposite answers. Git already draws that line: a documentation
directory is tracked by the commit that adds it, which is before CI and before review, so a reading
scoped to what git tracks meets every case the guard is for and none of the case it is not. It also
makes the local answer and the CI answer the same, which `DocumentationCorpus`'s own remarks say the
rooting exists to achieve.

Nothing historical is missed: every documentation directory in the repository is tracked.

The `.gitignore` reading goes with the filesystem walk — an ignored directory is untracked by
construction. `IgnoredRoots` stays, because `WorkflowTriggerCoverageTests` reads it.

The listing is handed in rather than read in the corpus, because reading it is what the fixture's own
`TrackedFiles` already does — with the `safe.directory` argument a checkout the process does not own
needs, which is every checkout the Unity job runs in. Reading git a second way got exit 128 there,
and the second way would have had to grow the same handling the first one already has cases for.

Both callers assert the listing arrived alongside what they conclude from it, because a reading that
answered nothing names no unwalked root either — the same silence a healthy corpus produces.

The case writes an untracked directory holding a document and asserts no root is named.

No documentation impact: the guard's failure message is its own text, and it now says to leave a
machine-local directory untracked rather than to add it to `.gitignore`.

Closes #802


s4k10503:
Verified RED by hand rather than by the base lane. Restoring the filesystem walk and re-running `DocumentationDriftTests` fails the new case; the fix passes it. The lane cannot separate them because `DocumentationCorpus` is a test-side file the base tree carries along with the cases.

PR #842, opened 2026-08-28
tooling(velvet): render a type nested in a generic as itself, not as its owner

### LINKED ISSUE #760: PublicApiSurfaceTests renders a nested type inside a public generic as its owner, so a change to either reads the same
`PublicApiSurfaceTests` renders a nested type inside a public generic identically to its owner, so it cannot tell an addition or removal on one from the other. The `VelvetTask` migration adds the repository's first such type, which is how this surfaced.

## Measured

`PublicAPI.txt` on that branch carries `[Velvet] type Velvet.VelvetTask\`1[T]` **twice** — the only duplicated line in the file — because `VelvetTask<T>` and its nested `Awaiter` render the same.

The consequence is not only a duplicate line. The Awaiter's members are recorded as the **outer type's**:

```
Velvet.VelvetTask`1[T].GetResult(): T
Velvet.VelvetTask`1[T].OnCompleted(System.Action)
Velvet.VelvetTask`1[T].IsCompleted
```

`VelvetTask<T>` declares none of those; all three belong to `Awaiter`. And `GetAwaiter()`'s return type renders mangled, as `Velvet.VelvetTask\`1[T~]`.

The **non-generic** case is correct: `Velvet.VelvetTask+Awaiter` renders with its owner named. So the defect is specific to a nested type inside a *generic* public type.

## What the guard stops answering

`PublicApiSurfaceTests` exists to make a change to the public surface visible in a diff a reviewer reads. For this shape it cannot:

- adding a member to `VelvetTask<T>.Awaiter` and adding one to `VelvetTask<T>` produce the same line;
- removing one from either produces the same deletion;
- and the duplicated type line means the file no longer round-trips its own input.

A reviewer reading `PublicAPI.txt` for what a major release changes is reading a rendering that has silently merged two types.

## Why it has not bitten before

No public nested type inside a public generic existed in this repository until now. The renderer's generic-arity handling (`\`1[T]`) and its nesting handling (`+Awaiter`) are correct separately and compose wrongly, which is the shape that survives until the first case reaches both at once.

## Scope

Fix the renderer so a nested type inside a generic owner names its owner the way the non-generic case does, and so a nested type's members are attributed to it. Then regenerate `PublicAPI.txt` on whatever branch is carrying `VelvetTask` at the time — the file is generated, so the regeneration is the evidence.

A case pinning the shape belongs with it: the fixture should hold a public generic with a public nested type and assert the rendering names both, since that is the composition nothing exercised.

Found while removing an unused public member from the `VelvetTask` migration; the hole is the guard's, not that branch's.

s4k10503:
`PublicApiSurfaceTests.FormatType` renders a type nested inside a public generic identically to its
owner, so the guard cannot tell an addition or a removal on one from the other.

`definition.FullName` for such a type is `Owner`1+Nested`, and the code cut at the **first** backtick —
which is the owner's arity marker — so `+Nested` was dropped and the nested type came out as the
owner. The arity appended after that is read off the type, and a type nested in a generic inherits its
owner's parameters, so the two rendered the same line. The non-generic case was already correct,
rendering `Owner+Nested`, which is why nothing reported it.

Measured on the `VelvetTask` branch, which adds the repository's first such type: `PublicAPI.txt`
carries `[Velvet] type Velvet.VelvetTask`1[T]` twice — the only duplicated line in the file — and the
`Awaiter`'s three members are recorded against `VelvetTask<T>`, which declares none of them.

Every arity marker is stripped now rather than the text before the first one taken. Nothing changes
for a type that is not nested: `Velvet.Foo`1` loses its marker and gets one back with the arguments,
as before.

The two cases assert the property the guard needs — the owner and the nested type render differently,
and the nested rendering carries the owner's name with `+Nested` behind it — rather than a spelling of
`FullName` written down here. What the marker looks like is read off the owner's own rendering, so
neither case claims anything about how the runtime spells one.

No documentation impact: `PublicAPI.txt` is generated, and this changes no line in it — the repository
has no type of this shape until #377 lands.

Closes #760

### DIFF
--- a/Packages/com.velvet.core/Runtime/Component/Tests/Editor/PublicApiSurfaceTests.cs
+++ b/Packages/com.velvet.core/Runtime/Component/Tests/Editor/PublicApiSurfaceTests.cs
@@ -107,6 +107,9 @@ namespace Velvet.Tests
             }
         }
 
+        // GREEN_ON_BASE(characterization): the rendering under test sits in this same file, which
+        // the base lane carries with the cases, so no base run can separate them. Restoring the
+        // first-backtick cut by hand and re-running the fixture fails both, which is the reading.
         [Test]
         public void Given_ATypeNestedInsideAGeneric_When_Rendered_Then_ItIsNotItsOwner()
         {
@@ -122,6 +125,9 @@ namespace Velvet.Tests
                         Is.EqualTo((true, true, true)));
         }
 
+        // GREEN_ON_BASE(characterization): the rendering under test sits in this same file, which
+        // the base lane carries with the cases, so no base run can separate them. Restoring the
+        // first-backtick cut by hand and re-running the fixture fails both, which is the reading.
         [Test]
         public void Given_ATypeNestedInsideAGeneric_When_Rendered_Then_ItsOwnerIsNamedInIt()
         {
--- a/Packages/com.velvet.core/Runtime/Component/Tests/Editor/PublicApiSurfaceTests.cs
+++ b/Packages/com.velvet.core/Runtime/Component/Tests/Editor/PublicApiSurfaceTests.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
+using System.Text.RegularExpressions;
 using NUnit.Framework;
 using Velvet.Editor.DevTools;
 using Velvet.TestUtilities;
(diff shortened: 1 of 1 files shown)


s4k10503:
Verified RED by hand rather than by the base lane. Restoring the first-backtick cut and re-running `PublicApiSurfaceTests` fails both cases; the fix passes them. The lane cannot separate the two because `PublicApiSurface` lives in the fixture's own file, which the base tree carries along with the cases.

PR #844, opened 2026-08-28
tooling(velvet): read a dated CHANGELOG section as the history it is

### LINKED ISSUE #799: A published release note pins every name it cites, so removing an API fails the drift guard
A dated CHANGELOG section describes a tree that shipped. `DocumentationDriftTests` reads it as a claim
about the tree that exists, so every name a published release note cites is pinned in the source
forever — and removing that name fails the guard.

## Measured

PR #377 removes UniTask from the package. `Packages/com.velvet.core/CHANGELOG.md`'s
`## [3.0.0] - 2026-08-27` section — published, tagged, and describing what v3.0.0 actually shipped —
cites the type twice. The EditMode run reports:

```
Expected: <empty>
But was:  < "Packages/com.velvet.core/CHANGELOG.md: UniTask (in `await UniTask.CompletedTask`)",
            "Packages/com.velvet.core/CHANGELOG.md: UniTask (in `async UniTask`)" >
```

Both citations are correct. v3.0.0 shipped `UniTask`, so its note names `UniTask`; a later release
renaming the type does not make the older note wrong, and rewriting a published section to say
`VelvetTask` would make it describe a version that never existed.

`IdentifierAllowlist` carries only the CHANGELOG's structural words — `Unreleased`, `Highlights`,
`Added`, `Changed`, `Breaking`, `YYYY`, `MM`, `DD`. Nothing exempts a dated section, so the walk treats
`## [3.0.0] - 2026-08-27` exactly as it treats `## [Unreleased]`.

## Why this is not #377's problem to solve

The two available repairs are both wrong:

- **Edit the released section.** `changelog_into_closed_version.py` refuses that write, and the note is
  already published under the tag, so the file and the release would disagree.
- **Allowlist `UniTask`.** That is one name for one removal. The next removal adds another, and the
  allowlist stops being a list of things the guard should not read and becomes a list of things the
  repository has deleted.

## The shape a fix wants

A section whose heading carries a date is a historical record. Its names resolved when it shipped, and
nothing in the current tree has to hold them. The `## [Unreleased]` and `## [Unreleased — breaking]`
sections are the opposite — they describe the tree being built, and a dangling name in one is exactly
what the guard exists to catch.

So the reading is per section, decidable from the heading, and it narrows rather than widens what the
guard asks. Whether it should also apply to the `### Highlights` block inside a dated section is the
same question one level in.

**Measure the yield before building it**: over this repository's CHANGELOG, how many names in dated
sections resolve today, and how many would stop being asked. A narrowing that silences a real dangling
reference in an open section would be worse than the defect.

Blocks #377.

s4k10503:
A dated CHANGELOG section describes a tree that shipped. `DocumentationDriftTests` read it as a claim
about the tree that exists, so every name a published release note cites was pinned in the source
forever — and removing that name failed the guard.

Measured on the branch that removes UniTask: `## [3.0.0] - 2026-08-27` is published, tagged, and cites
the type twice, so the EditMode run reported both citations. Both are correct. v3.0.0 shipped
`UniTask`, so its note names `UniTask`, and a later release renaming the type does not make the older
note wrong.

Neither available repair works. Editing the released section is refused by
`changelog_into_closed_version.py`, and the note is already published under its tag, so the file and
the release would disagree. Allowlisting the name is one name for one removal — the next removal adds
another, and the allowlist stops being a list of what the guard should not read and becomes a list of
what the repository has deleted.

So the reading is per section and decidable from the heading, and it narrows what the guard asks
rather than widening it. `## [Unreleased]` and `## [Unreleased — breaking]` keep their reading, which
is the one the guard exists for: they describe the tree being built, and a dangling name in one is
drift. Measured against this repository's own CHANGELOG, 2480 lines come out as 508, and the headings
that survive are exactly those two.

A `### Highlights` block inside a dated section goes with it, which is the same question one level in
— the range runs to the next version heading. A guide that happens to spell a heading the same way is
left alone: only the CHANGELOG carries the convention.

Five allowlist entries went dead with the narrowing and are removed with it: `Save`,
`NullReferenceException`, `BringToFront`, `SendToBack` and `SetCursor`. Every one was written only by
a dated section, which is the shape the allowlist was being asked to carry — one name per removal, a
list of what the repository has deleted. `Given_TheIdentifierAllowlist_..._Then_EveryEntrySuppressesAReport`
is what named them, which is the sibling guard doing its job.

The three cases read `WithoutShippedSections` directly and are declared, because it is declared in the
same file the base lane carries with them — no base run can separate the two. What pins the wiring is
the guard itself: taking the call out of `BacktickSpans` and re-running the fixture reddens
`Given_ProjectMarkdown_When_ScannedForBacktickedIdentifiers_Then_EveryIdentifierAppearsInASource`,
measured, which is the case this change exists for.

No documentation impact: the guard's failure message is its own text.

Closes #799


s4k10503:
Verified by hand, and the first attempt measured the wrong thing. The three new cases call `WithoutShippedSections` directly, so removing its call site in `BacktickSpans` leaves them green — they pin the helper, not the wiring. What the wiring is pinned by is `Given_ProjectMarkdown_When_ScannedForBacktickedIdentifiers_Then_EveryIdentifierAppearsInASource`, which reddens without the call. That is the reading, and it is in the body now.

PR #868, opened 2026-08-29
test(velvet): drop the hop the suspend resolution does not take

### LINKED ISSUE #585: RouterTests' post-commit Suspend-failure case fails on CI and never locally
`Given_SuspendLoaderFailedAfterCommit_When_GoBackToIt_Then_RestoresCachedError` (`Runtime/Routing/Tests/Editor/RouterTests.cs`) failed one EditMode run on a branch that touches only `.claude/` — 3783/3784, this the only failure. The same fixture run three times on a developer machine passed 50/50 each time, so what separates the runs is the runner, not the change.

The stack in the failing run points inside the `UniTask.ToCoroutine` lambda, around the `LogAssert.Expect` / `await UniTask.Yield()` / `Assume.That(CurrentLoaderErrors.Count, Is.EqualTo(1))` sequence — the test sets the loader's exception, yields **once**, and then requires that the failure has already been recorded.

That single yield is the suspect and it is the shape this repository has been bitten by before: a wait sized to how long something took on the machine that wrote it. A CI runner has no GPU and a contended scheduler, so one continuation hop is not the same amount of progress there.

## What would close it

Not a longer sleep and not a retry. Either the recording is synchronous with the exception being set — in which case the yield is the bug and removing it makes the test deterministic and the production ordering explicit — or it genuinely takes a scheduler hop, in which case the test must wait on the *condition* through whatever the routing layer offers as a completion signal rather than on a hop count.

Establish which by reading `OnSuspendLoaderFailed`'s path from `UniTaskCompletionSource.TrySetException` to `CurrentLoaderErrors`, not by adding yields until it stops failing.

Whatever the fix, it applies to every test in the repository that awaits a fixed number of hops after arming an async primitive — that set should be enumerated as part of closing this, because one fixed-hop wait that survives is the next intermittent failure.

Evidence: run 31325811571, job 'Unity tests (EditMode)', on `tooling/changelog-closed-section` (a `.claude/`-only branch). A rerun of the same job on the same SHA is the confirmation that it is not the change.

s4k10503:
`Given_SuspendLoaderFailedAfterCommit_When_GoBackToIt_Then_RestoresCachedError` waited one
continuation hop between setting the loader's exception and requiring the failure to be recorded. One
hop is not the same amount of progress on a CI runner as on the machine that counted it, which is how
the case failed once there and never locally.

The recording does not need a hop. `NavigateSync` drives `RunSuspendLoader` to its `await task` before
it returns, so the continuation is registered by the time the test sets the exception;
`UniTaskCompletionSource.TrySetException` resumes it inline; the catch reaches `Announce`, which is a
direct `subscribers?.Invoke`; and the Router's `OnSuspendLoaderFailed` handler writes
`_loaderErrors` inline. Nothing on that path defers.

So the wait goes, and the `Assume` beside it becomes the thing that says so — with the hop removed it
is red the moment any step on that path stops being inline. Measured: adding a single
`await UniTask.Yield()` before `Announce` in `RouteLoaderRunner.RunSuspendLoader` fails this case and
no other; the fixture is 51/51 with it removed, over three runs.

Its sibling `Given_SuspendLoaderResolvedAfterCommit_When_GoBackToIt_Then_RestoresCachedData` had the
same shape on the completion path and is the same change.

Those two are the whole set the issue asks to enumerate. Of the 38 `await UniTask.Yield()` calls in
the package, across three files, exactly two are followed by an assertion — a wait before an
unsubscribe or before a further act is a step in the arrangement, not a claim about how long something
takes. The one in
`Given_ASubscriberThatThrowsOnASuspendResolution_When_GoBackToThatEntry_Then_TheCacheStillServesIt`
stays for that reason: what follows it is `router.OnLocationChanged -= Throwing`, and the order of
those two is what the case is about.

Closes #585

### DIFF
--- a/Packages/com.velvet.core/Runtime/Routing/Tests/Editor/RouterTests.cs
+++ b/Packages/com.velvet.core/Runtime/Routing/Tests/Editor/RouterTests.cs
@@ -555,8 +555,9 @@ namespace Velvet.Tests
             router.NavigateSync("/deferred");
             Assume.That(router.GetLoaderData("/deferred"), Is.Null, "Precondition: unresolved at commit time");
             tcs.TrySetResult("deferred-data");
-            await UniTask.Yield();
-            Assume.That(router.GetLoaderData("/deferred"), Is.EqualTo("deferred-data"), "Precondition: resolved after commit");
+            // No hop between the two, for the reason the failure case below gives.
+            Assume.That(router.GetLoaderData("/deferred"), Is.EqualTo("deferred-data"),
+                "Precondition: resolved synchronously with the result being set");
             router.NavigateSync("/other");
 
             // Act
@@ -586,8 +587,12 @@ namespace Velvet.Tests
             // Suspend-mode failures route through OnSuspendLoaderFailed, which logs the exception.
             LogAssert.Expect(UnityEngine.LogType.Exception, new System.Text.RegularExpressions.Regex("deferred-failure"));
             tcs.TrySetException(new InvalidOperationException("deferred-failure"));
-            await UniTask.Yield();
-            Assume.That(router.CurrentLoaderErrors.Count, Is.EqualTo(1), "Precondition: the failure was recorded after commit");
+            // No hop between the two: the completion source resumes the loader's `await` inline, the
+            // catch announces inline, and the Router's handler writes the error inline. A wait here
+            // would let a future hop through unnoticed, and one hop is not the same amount of progress
+            // on a runner as on the machine that counted it.
+            Assume.That(router.CurrentLoaderErrors.Count, Is.EqualTo(1),
+                "Precondition: the failure was recorded synchronously with the exception being set");
             router.NavigateSync("/other");
 
             // Act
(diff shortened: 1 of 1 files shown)


s4k10503:
Two corrections to this description, both mine.

**The CI failure it cites did not happen.** Run 31325811571 reports one failure and it is
`PreExpansionPolicyTests.Given_EveryRefusingGuard_When_ItsSourceIsRead_Then_ItStatesAPreExpansionPolicy`;
`Given_SuspendLoaderFailedAfterCommit_When_GoBackToIt_Then_RestoresCachedError` is `result="Passed"` in
the same results file. I took the premise from #585 without opening the run. What justifies the change
is only the second half of the description — the path is inline, so the wait separated nothing, and
removing it makes the case red if that ever stops being true.

**"Exactly two of the 38 are followed by an assertion" needed its window stated.** The two are the
ones whose very next statement is an `Assume`; read over a whole test method, most yields have an
assertion somewhere after them, because most tests end in one. On `main` after this merge there are 36
`await UniTask.Yield()` calls and none has an assertion as its next statement, which is the same
measurement from the other side: the two that did are the two this removed.

PR #875, opened 2026-08-29
fix(velvet): give the guard a flag for every slot a mode drives

### LINKED ISSUE #580: animate-hue and the gradient pair have no slot flag, so a transition over their longhand intercepts every tick
> **Re-measured against `origin/main` (`89042a0`) on 2026-08-13.** The fix this issue asked for —
> routing the animate driver through the existing guard rather than writing a second mechanism —
> landed for two of the five modes. The headline example (`transition-transform animate-spin`) no
> longer reproduces. What remains is the three modes whose slots the guard has no flag for.

`StyleAnimateDriver` writes an inline value every tick. When a transition covers that same longhand,
UI Toolkit does not write the value — it starts a fresh native transition toward it. The painted
value then trails the driver for the whole loop.

A transition needs no `transition-*` utility to be present: a bare `duration-*` leaves UI Toolkit's
initial `all` standing, which covers every slot.

## What is closed

`StyleAnimateDriver` now takes the guard. `SyncTransitionSuspension` is called on attach
(`Runtime/Styling/StyleAnimateDriver.cs:64`), forwards to
`MotionNativeTransitionGuard.SyncSuspension` (`:77-78`), and releases on detach (`:90`).

That covers `animate-spin` (rotate) and `animate-pulse` (opacity), including the original repro:
`className="transition-transform animate-spin"` suspends the transition for the length of the spin.
Pinned by `AnimateMotionPanelTests.cs:240-257` and `AnimateSuspensionUnderVariantSwapTests.cs:116-142`.

## What remains

`animate-hue` and the gradient pair are still unguarded, and the reason is structural rather than an
oversight at the call site: `MotionTransitionSlots`
(`Runtime/Styling/MotionNativeTransitionGuard.cs:12-26`) declares
`None | Opacity | Translate | Scale | Rotate | Color | Length` and has no member for `filter` or
`background-position`. So `GuardedSlots` (`Runtime/Styling/StyleAnimateDriver.cs:82-87`) returns
`MotionTransitionSlots.None` for Hue, Gradient and Shimmer, and `SyncSuspension` short-circuits to
`Release` on `None`.

`animate-hue` under anything covering `filter`, and the gradient pair under anything covering
`background-position`, therefore still trail the driver for the whole loop.

## What would close it

Add the two slot members and their `SlotsOf` mapping. Measure first which longhands each mode
actually writes — the gradient pair drives `background-position`, `background-size` and
`background-repeat`, and only the first is a plausible transition target.

`Documentation~/motion.md:333-340` states the current split in as many words ("`animate-hue` and the
gradient pair are not: their slots have no flag in the guard yet") and must be updated with the fix.

s4k10503:
`MotionNativeTransitionGuard` suspends an element's native USS transitions while a per-frame driver
writes a slot they would otherwise intercept. `MotionTransitionSlots` had a member for opacity,
translate, scale, rotate, colour and length — and none for `filter` or background position. So
`animate-hue` and the two gradient pan modes reported driving nothing, `SyncSuspension`
short-circuited to `Release`, and a transition covering those longhands took every tick's write.

A bare `duration-*` is enough to reach it: UI Toolkit's initial `transition-property` of `all` covers
every slot, so no `transition-*` utility has to be present. That is also how an element reaches the
two new slots, because no bundled utility names either of them on its own.

Measured which longhand each mode writes per tick, since only a per-tick write is what a transition
intercepts frame after frame: Hue writes `filter`; Gradient and Shimmer write one of
`backgroundPositionX` / `backgroundPositionY`, whichever the binding's own axis names. Both new slots
follow Colour's precedent of standing for the group rather than the channel.

What the suspension does not reach is the background size and repeat the pan modes set at attach.
That write happens once, and above the call that takes the suspension, so a transition covering
either still takes it. The guide says so rather than claiming the modes are now uniformly safe.

The declared side is deliberately left alone here. `SlotsOf` gains no arm for either slot, so the
cascade oracle's reading is unchanged and no carried test module names a member the base has not got
— which is what lets `base-red` answer this in one round rather than none. `.transition-filter` is
the one bundled utility that could name a declared side finer, and giving it an arm is a change that
can be read on a base that already has the member.

Two cases, one per new slot, each red without its `GuardedSlots` arm and red on the base.
EditMode 4196 passed / 0 failed, PlayMode 161 / 0.

Closes #580

### DIFF
--- a/Packages/com.velvet.core/CHANGELOG.md
+++ b/Packages/com.velvet.core/CHANGELOG.md
@@ -59,7 +59,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the element's own transitionable set, and it had no flag for `filter` or for the background-position
   pair — so those three reported driving nothing and their per-frame writes were left to whatever
   transition covered them. A bare `duration-*` is enough to reach this, because UI Toolkit's initial
-  `transition-property` of `all` covers every slot. What the suspension does not reach is the
+  `transition-property` of `all` covers every slot, which is how an element reaches the two new slots:
+  no bundled utility names either of them on its own. What the suspension does not reach is the
   background size and repeat the pan modes set once at attach, above the call that takes it.
 
 - A component inside one `V.Portal` no longer has its position rewritten by a sibling portal on a
--- a/Packages/com.velvet.core/CHANGELOG.md
+++ b/Packages/com.velvet.core/CHANGELOG.md
@@ -57,11 +57,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `animate-hue` and the two gradient pan modes now suspend a native transition covering the slot they
   drive, as `animate-spin` and `animate-pulse` already did. The guard matches a driver's slot against
   the element's own transitionable set, and it had no flag for `filter` or for the background-position
-  pair — so those three reported driving nothing and were left to whatever transition covered them.
-  UI Toolkit does not take an inline write while a transition covers that longhand; it starts a fresh
-  transition toward it, so the painted value trailed the driver for the whole loop. A bare `duration-*`
-  is enough to reach this, because UI Toolkit's initial `transition-property` of `all` covers
-  every slot.
+  pair — so those three reported driving nothing and their per-frame writes were left to whatever
+  transition covered them. A bare `duration-*` is enough to reach this, because UI Toolkit's initial
+  `transition-property` of `all` covers every slot. What the suspension does not reach is the
+  background size and repeat the pan modes set once at attach, above the call that takes it.
 
 - A component inside one `V.Portal` no longer has its position rewritten by a sibling portal on a
   different target. Two portals declared side by side are siblings in the fiber chain whatever targets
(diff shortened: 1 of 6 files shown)


s4k10503:
`Base-red check (EditMode)` reports both cases as `the base tree cannot answer`. That is the
single-round reading, and it is not the answer: the cases name `MotionTransitionSlots.Filter`, which
the base has no member for, so the carried module does not compile there and the round writes nothing.

Taken where the loop runs, on the same base with a warm Library:

```
EditMode attempt 1:  0 case(s) over 4 fixture(s) in 15s
EditMode attempt 2: 28 case(s) over 4 fixture(s) in 18s

red on the base   Velvet.Tests.AnimateMotionPanelTests.Given_AHueUnderADuration_...
red on the base   Velvet.Tests.AnimateMotionPanelTests.Given_AGradientPanUnderADuration_...
```

Two rounds: the first withdraws the module the base cannot build, the second measures. Both cases are
red on the base, which is what they are for.

This is the divergence #635 records — the local loop withdraws and the single CI round cannot — and
the refusal names that command itself.

PR #878, opened 2026-08-29
feat(velvet): ship an awaitable of our own, before anything returns one

s4k10503:
The first of two. This adds `VelvetTask` / `VelvetTask<T>` and nothing returns or takes one yet; the
change moving Velvet's own signatures onto it follows, against a tree that already has the type.

Split for a reason a reviewer can check rather than for size. `base_red_check` builds the base by
copying the branch's changed test modules onto the merge base, so a change that alters a signature's
type makes every carried module naming it uncompilable there — and #377, which does that across nine
assemblies at once, gets no verdict for any of its 280 cases. Additive is the half that can be read:
the carried modules name a type the base has not got, which the static withdrawal is written for, and
what remains measures.

`VelvetTask` follows UniTask's design, with the surface this codebase uses and no more: `Forget`,
`FromResult`, `AttachExternalCancellation`, completion sources, `Yield`, `Never`, `FromException`,
`CompletedTask`, `ToCoroutine`, and a synchronous status peek. No `WhenAll`, no `Delay`, no async
LINQ — those have no call sites here. A pooled source carrying a version token, a custom async method
builder that stays lazy until first suspension, and completed tasks held inline in the struct, so a
synchronously-completed path allocates nothing.

Frame-bound continuations are driven from the PlayerLoop at runtime and from `EditorApplication.update`
in the editor, which is what lets an `async VelvetTask` method resume in EditMode — where most of this
package's suites run. A `UnityEngine.Awaitable` and a BCL `Task` need no adapter: both builders declare
the two awaiter constraints the compiler emits an await against, so either is awaited inside an
`async VelvetTask` as it stands.

Two test assemblies come with it, 44 cases across EditMode and PlayMode. `PublicAPI.txt` and
`AssemblyGraph.txt` are regenerated. `TestUtilities` gains the frame-driver helper those suites drive
continuations with, and the two `InternalsVisibleTo` grants that reach it.

EditMode 4240 passed / 0 failed, PlayMode 184 / 0.

Refs #377 — this is its first half; the migration stays there.


s4k10503:
`Duplication ratchet` reports 15 blocks that did not repeat before, and 14 of them are between
`VelvetTaskSources.cs` and `VelvetTaskStateMachineRunners.cs`. They are the void/`T` pair:
`AsyncVelvetTaskMethod<TStateMachine>` beside `AsyncVelvetTaskMethod<TStateMachine, T>`, and the same
split one level down in the sources. C# has no way to write one of those that serves both without
boxing the result or inventing a unit type to stand where `void` does, which is why UniTask ships the
same pair for the same reason and why this follows it.

The fifteenth is a subtraction, not an addition: `Runtime/AssemblyInfo.cs` and
`TestUtilities/AssemblyInfo.cs` stop matching, because the grants for the two new test assemblies
land in one of them.

Baseline updated, which is what the ratchet asks for when the duplication is deliberate.

PR #928, opened 2026-08-30
fix(velvet): mutate the lines the branch wrote, not the ones a diff calls changed

### LINKED ISSUE #684: mutation_check attributes lines the branch never wrote, where base_red_check no longer does
`base_red_check.py` and `mutation_check.py` each carry their own `changed_lines_by_file`, and both read git's line-attributed diff. Over a heavily rewritten file the aligner describes an untouched tail as deleted and re-added, so both attribute lines the branch did not write.

## Measured

On `test_base_red_check.py` alone, comparing the attributed case set against each case's own text on both sides: **11 extra cases**, every one byte-identical on both trees, all reported `passed on the base`.

PR #634 fixes this for `base_red_check.py` by comparing each case's own text rather than trusting the line attribution — 0 spurious, 0 lost. **`mutation_check.py` was deliberately left alone** and now differs from its sibling.

## What it costs there

A mutation campaign over a heavily-rewritten file mutates lines the branch did not write, and reports survivors for code somebody else owns. The author then either answers a `MUTANT_SURVIVES` declaration for a line they never touched, or the receipt is refused over one. Neither is the question the campaign exists to ask.

## Why it was not fixed in #634

Two reasons, both worth carrying forward:

- `mutation_check.py` has a 113-case suite and a committed baseline, so a change there has a much larger blast radius than the same change in its sibling.
- **The obvious alternative could not be pinned.** `--patience` / `--histogram` / `--minimal` also give 0 spurious and 0 lost on the measured corpus, but no repro was found that reddens without them — the move-a-class repro that was built is attributed identically by all four algorithms. An unpinnable change was correctly declined rather than landed.

So the text-comparison approach #634 took is the one with evidence behind it, and porting it is the tractable path. Whoever takes this should start by reproducing the 11 on the current tree, since the number moves as the file does.

## Related

The same reading feeds `--emit`'s case selection, so a fix here also narrows what a campaign is asked to cover, which is the direction the receipt gate wants anyway.

s4k10503:
Closes #684.

`changed_files_and_lines` trusted git's line attribution, and which lines a diff calls changed is
git's choice rather than the file's: over a rewrite the aligner is free to describe an untouched run
as deleted and re-added, and every line in it then reads as this branch's. A campaign mutates code
somebody else owns, and the author either answers a `MUTANT_SURVIVES` declaration for a line they
never wrote or has the receipt refused over one. `base_red_check.py` stopped believing the same
attribution in #634; this is its sibling reading, which was deliberately left alone then.

Measured over the sixty most recent commits on `main`, comparing each attributed line's content
against the merge base:

    attributed lines                                     2977
    content the base already held as many copies of        42

They are not spread evenly. `RouteBlockerManager.cs` at c19be244d carries 16 of its 57 — a whole
`if`/`else if` chain that stayed identical while the method around it was restructured — and
`shell_commands.py` at 190abb9d0 carries 10 of 26.

The comparison is a multiset over the file rather than position by position, so a line the branch
genuinely added survives even where the base holds one like it; that is the second case. What it
cannot see is a line moved into a different context, where the move is the behaviour — the same blind
spot `authored` has at its own granularity, and the docstring says so rather than implying coverage.

The red case is the shape those sixteen took: statements wrapped in a new construct, so every one is
re-indented and the diff calls all of them added while the branch wrote none of their behaviour.
Measured red on the base; the control is declared `GREEN_ON_BASE(characterization)`.

No receipt is invalidated. `scope_digest` hashes `(base, platform, [relative path : sha256 of the
file])` — the files a campaign covered, not the lines within them — and every file the diff names
stays a key here whether or not any of its lines survive the narrowing.

Suite 187 passed / 0 failed.


s4k10503:
Withdrawing this. A review found the narrowing silences a line the branch genuinely wrote, and I
reproduced it: a guard removed from one method and written into another, where it is new behaviour.

    base    class A { void M() { if (a) { return; } }  void N() { var x = 1; } }
    branch  class A { void M() { }  void N() { var x = 1; if (a) { return; } } }

    on main         attributed: ['if (a) { return; }']
    on this branch  attributed: []

No mutant is generated for it, and a campaign then reports nothing survived about a line it never
touched — which is the failure this repository treats as the worst kind, because it is
indistinguishable from a mutant that was generated and killed. That is the direction the change was
supposed to be safe in.

The obvious tightening does not pay for itself. Requiring a line to sit in a run of three or more
balanced lines keeps the safe direction, and I measured what it costs over the sixty most recent
commits on `main`:

    attributed lines                                     3071
    balanced against the base (this branch's rule)         41
    of those, in a run of three or more                    17

`RouteBlockerManager.cs` at c19be244d — the case this was built for — keeps only 5 of its 16 under the
run rule, because the aligner matched the block's braces as context and called only the statements
between them re-added, so the balanced lines are separated by unchanged ones. So the run rule recovers
0.55% of attributed lines, and still silences a whole block moved wholesale, which is the same defect
three lines up.

Recorded on #684 with the numbers. The short of it: the text comparison #634 took works at case
granularity because a case has a name and its own code, and lines have neither — so the port is not
the tractable path the issue took it for.

PR #965, opened 2026-09-05
tooling(velvet): refuse a -testFilter naming a class no test file declares

### LINKED ISSUE #962: A -testFilter derived from filenames misses 19% of fixture classes, and the run reports green over the smaller set
## Measured

A `-testFilter` built from the names of the files a change touched under-selects, because a test file may declare more than one fixture class. Over `Packages/com.velvet.core`:

| | |
|---|---|
| test files declaring a `*Tests` class | 311 |
| fixture classes declared | 383 |
| files declaring more than one | 44 (14%) |
| files whose stem names no class at all | 7 |
| **classes a filename-derived filter misses** | **72 (19%)** |

Worst cases: `WidgetPoolHelperTests.cs` declares five (`ButtonPoolHelperTests`, `LabelPoolHelperTests`, `SliderPoolHelperTests`, `TextFieldPoolHelperTests`, `TogglePoolHelperTests`); `GapParityTests.cs` and `ResponsiveBreakpointPanelTests.cs` declare three each.

## What it costs

A filtered run reports a green count for a set smaller than the one the author believed they measured, and nothing says so. `assert_results_from_this_tree.py` answers a different question — whether the results came from this worktree — and `assert_no_inconclusive.py` a third. Neither compares what the filter asked for against what ran.

Both halves were hit in one session, independently:

- A run over a 40-name filter reported 37 distinct fixture classnames. The three that matched nothing were not chased, so the run's own count carried an unexamined gap.
- A routing run's filter named `RoutingHooksTests` from the filename; the file also declares `RouteNavigationStateTests` and `RouteNavigateHookTests`, which hold the two cases the change under test was about. The run was green over a set that excluded them.

## A decidable rule

Naming *too few* classes is not decidable in general — an author may filter to one fixture on purpose. But the shape both failures took is decidable: **a filter names a class whose declaring file declares other fixture classes the filter does not name.** That is exactly the 44-file, 72-class exposure above, and it is answerable by parsing the file the named class is declared in.

The other half is decidable outright: **a filter names a class no file declares** — a typo, or a name a rename left behind.

A refusal would be wrong for the first (filtering to one fixture is legitimate); printing the sibling classes it did not name is not. The second can refuse.

## Where

- A `PreToolUse` hook on a Bash command carrying `-testFilter`, alongside the existing `.claude/hooks/refuse/` family.
- Or a small `scripts/test_quality/` helper that prints the correct filter for a set of files, so the derivation stops being done by hand — the failure is in deriving the filter from filenames, not in the filter mechanism.

CONTRIBUTING.md and CLAUDE.md both document `-testFilter` as "matches fully-qualified class or method names" without saying that a file's name is not one.

s4k10503:
Closes #962.

A `-testFilter` written from the names of the files a change touched selects fewer fixtures than its
author believes, because a test file may declare more than one. Measured over this project: 323 of its
349 test files declare a fixture class, 401 classes between them; 46 of those files declare more than
one and 10 declare none matching their own stem, which leaves 88 of the 401 classes invisible to a
filename-derived filter. Neither way it goes wrong is loud — the run comes back green over a smaller
set, and the count reads the same as one taken over the whole file.

It has cost coverage here. A routing change's filter named `RoutingHooksTests` from the filename; the
same file declares `RouteNavigationStateTests` and `RouteNavigateHookTests`, and one of the three cases
that killed that change's mutant lives in the first of those. The green count was reported over a set
that excluded the fixture the change was about.

The new `PreToolUse` guard reads the fixture classes the project's test sources declare and answers two
questions about a filter, at two severities. A name nothing declares selects none of the cases it asks
for, so it is refused, with the classes of the file it was probably named for printed beside it. A name
that leaves other classes of its own file out is how a deliberate single-fixture run is written, so
those classes are named in a `systemMessage` and the command goes through.

The split is measured rather than assumed. Replayed over the 1709 distinct `-testFilter` commands this
project's transcripts hold, the notice fires on 90 of the 1523 whose filter is a readable literal, and
on 16 of those the filter already names more than one class of the same file — the author had read its
class list. Refusing there would have been noise on roughly one filtered run in a hundred.

The reading is two-stage so it can sit in front of every Bash command: a raw scan picks the files a name
could be declared in, and only those are masked and parsed. `base_red_check.py` owns how a case's
fixture name is read off C#, and is loaded rather than re-implemented.

`CLAUDE.md` and the unity-tests skill both documented the flag as matching fully-qualified class or
method names without saying that a file's name is neither; `CLAUDE.md` now carries that, and the skill
points at it.


s4k10503:
Closing without merging. The guard does not clear the bar this repository sets for one, and the measurement that says so is the branch's own.

Seven rounds, each finding a defect the previous round's fix created, and every one of them the same class: **the guard failing to fire, in a way indistinguishable from having nothing to say.** The last is the sharpest. Round 7 added a parse cache to fix a real timeout — the guard took 12.7–17.7 s against a 20 s registration, and a `PreToolUse` hook that overruns does not refuse, it goes quiet. The cache is keyed on the C# file's bytes and not on the parser that read them, in a `$TMPDIR` shared by every checkout and session on the machine. Measured in an isolated sandbox with two hooks differing only in `_parse`: whichever fills the entry first decides the answer for the other, so a hook whose own parser says refuse answers 0. And four of the guard's own cases are red on a cold cache and green on a warm one — the four whose subject is the parse — so on any machine that has run the suite once, a `_parse` regression ships green.

What it was worth, measured over 2100 command/cwd pairs from this project's transcripts: 59 refusals, 34 distinct values, **2 genuine**. The rest are replay artefacts — scratch probes and stems a `.cs` once had. Against the repository's own 4354 case names in three spellings, 1 refusal.

And the failure it uniquely catches is narrower than it looked. `scripts/test_quality/assert_results_from_this_tree.py` already refuses a filter that selects nothing at all. What it cannot see is one non-selecting value beside a selecting one — a two-value filter reports the tests the good half selected, and both asserters exit 0. That gap is real and it is one shape.

[#962](https://github.com/s4k10503/velvet/issues/962) stays open with the problem it records. What is true and reusable from this branch is the `-testFilter` semantics it established by decompiling Unity's test framework — the value is an unanchored .NET regex over the full test name, a leading `!` routes through `NotFilter`, and a type that only nests fixtures reports them as its siblings under the namespace rather than below it. Those are facts about the flag, not about the guard, and they belong in the places that describe the flag.

